PR #20, opened 2019-09-16
Convert #define to constexpr

### LINKED ISSUE #14: Convert #define to constexpr
The BusTub has a bunch of #define'd stuff. For example, there is [config.h](https://github.com/cmu-db/bustub/blob/master/src/include/common/config.h) and more generally you can just [search for define](https://github.com/cmu-db/bustub/search?q=define&unscoped_q=define). We would prefer these to be constexpr instead.

To quote Matt, that gives us "proper namespace-scoping, type-checking, clang-tidy naming checks, and nicer behavior with IDEs".

tupaschoal:
Closes #14


lmwnshn:
@tupaschoal Thanks for the PR. Is this still WIP?

tupaschoal:
It is not @lmwnshn . I put WIP to wait for the CI results.

PR #23, opened 2019-09-23
Fix the memory leak in the sample buffer pool test.

lmwnshn:
Google test is a bit of a pain. Even if you delete the disk manager, you'll leak memory. You need both disk manager and buffer pool manager to be pointers to be cleanly deleted.

We already fixed this in the autograder last/lastlast week, but we forgot to port it here. Hopefully the new infrastructure will fix this.

### DIFF
--- a/src/include/catalog/column.h
+++ b/src/include/catalog/column.h
@@ -99,7 +99,9 @@ class Column {
       case TypeId::VARCHAR:
         // TODO(Amadou): Confirm this.
         return 12;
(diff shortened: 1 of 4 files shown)


lmwnshn:
Not exactly.. there was a real leak, but even if you fixed that, there is still a bug in how we write our tests because of how Google tests work. The buffer pool manager is alive for too long, after the "end of the test", so it looks like it is a potential leak. That's my understanding, anyway..

PR #25, opened 2019-09-24
Apply naming conventions from Terrier

lmwnshn:
This brings in Matt's naming convention clang-tidy checks from Terrier and applies it to the existing codebase. It should not conflict with existing student files because it does not touch the clock replacer or buffer pool stuff. We will not enforce this check on current students; we will switch it back on before Fall 2020.

### DIFF
--- a/src/include/storage/table/table_heap.h
+++ b/src/include/storage/table/table_heap.h
@@ -108,12 +108,12 @@ class TableHeap {
   /**
    * Return the begin iterator of the table.
    */
-  TableIterator begin(Transaction *txn);
+  TableIterator Begin(Transaction *txn);
 
   /**
    * Return the end iterator of the table.
    */
-  TableIterator end();
+  TableIterator End();
 
   /**
    * Return the id of the first page of this table.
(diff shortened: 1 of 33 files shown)


src/include/storage/table/table_heap.h
```diff
@@ -108,12 +108,12 @@ class TableHeap {
   /**
    * Return the begin iterator of the table.
    */
-  TableIterator begin(Transaction *txn);
+  TableIterator Begin(Transaction *txn);
```
amlatyrngom: Needs to be lower case for std iteration.

lmwnshn: It looks like std::iterator was [deprecated](https://stackoverflow.com/questions/37031805/preparation-for-stditerator-being-deprecated/38103394) in C++17, actually. Perhaps it is OK to break this and open an issue to move to whatever standards are being dreamed up right now.

lmwnshn: No strong opinion.

lmwnshn: I'm just going to merge it for now.

PR #46, opened 2019-12-18
Build: Update cpplint for C++17 tuple destructing

at15:
The old version complains after format because format wants `auto [a,
b]` while lint think `auto[a, b]` is the right one i.e. no extra space. It is fixed in upstream already https://github.com/google/styleguide/pull/487

NOTE: I am using https://github.com/google/styleguide/tree/gh-pages/cpplint instead of the fork https://github.com/cpplint/cpplint The fix is in both repos and I am not sure the origin of the one in bustub repo, the diff with both repos are pretty large so I picked the google one for now.

### DIFF
--- a/build_support/cpplint.py
+++ b/build_support/cpplint.py
@@ -44,30 +44,48 @@ same line, but it is far from perfect (in either direction).
 import codecs
 import copy
 import getopt
+import glob
+import itertools
 import math  # for log
 import os
 import re
 import sre_compile
 import string
 import sys
-import unicodedata
 import sysconfig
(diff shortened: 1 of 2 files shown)


apavlo:
@lmwnshn Can you check this? We may want to copy it over to the the main DBMS.

lmwnshn:
@at15 Thanks! This is very useful. Could you please use the fork? I'll merge once that is in.

@apavlo Yup, looks good and we want this. I remember NOLINTing a bunch of auto's in the exec engine last semester.

PR #77, opened 2020-08-21
Cpp starter code for P0

preetansh:


src/include/primer/p0_starter.h
```diff
@@ -0,0 +1,119 @@
...
+  // Return the (i,j)th  matrix element
+  virtual T getElem(int i, int j) = 0;
+
+  // Sets the (i,j)th  matrix element to val
+  virtual void setElem(int i, int j, T val) = 0;
+
+  // Sets the matrix elements based on the array arr
+  virtual void matImport(T *arr) = 0;
+
+  // P0: Add implementation
```
lmwnshn, line 51: Nit: I think in the past, we've marked "students should implement this" as "TODO(P0): ...".

src/include/primer/p0_starter.h
```diff
@@ -0,0 +1,119 @@
...
+  ~RowMatrix() override = default;
+
+ private:
+  // 2D array containing the elements of the matrix in row-major format
+  // P0: Allocate the array of row pointers in the constructor. Use these pointers
+  // to point to corresponding elements of the 'linear' array.
+  // data should contain pointers to the rows in the linear array to make elements addressable using two dimensions.
+  // NOTE : Do not duplicate elements stored in linear.
+  // Don't forget to free up the array in the destructor.
+  T **data;
```
lmwnshn, line 86: Nit: I'm not sure if we have the same clang-tidy settings as terrier, but generally private members are postfixed with an underscore, `data_`.

src/include/primer/p0_starter.h
```diff
@@ -0,0 +1,119 @@
...
+  // # of Columns in the matrix
+  int cols;
+  // Flattened array containing the elements of the matrix
+  // P0: Allocate the array in the constructor. Don't forget to free up
+  // the array in the destructor.
+  T *linear;
+
+ public:
+  // Return the # of rows in the matrix
+  virtual int getRows() = 0;
```
lmwnshn, line 37: Nit: GetRows instead of getRows, etc?

PR #83, opened 2020-08-28
Remove nonportable and unused cmake code

hcnelson99:
The deleted cmake code is copied from a low-quality stack overflow
answer [0]. The answer suggests adding a CMAKE_CXX_FLAG in terms of $<,
which is a Makefile-ism. However, CMake is supposed to support other
build system targets such as Ninja. Referring to $< breaks the Ninja
build. It appears that the __BUSTUBFILE__ macro is unused in the code
base anyways so this patch just deletes the unused and buggy CMake code.
If this __BUSTUBFILE__ macro is desired, it should be implemented in
terms of portable CMake code such as in [1] instead.

[0] http://stackoverflow.com/a/16658858
[1] https://stackoverflow.com/questions/1706346/file-macro-manipulation-handling-at-compile-time/27990434#27990434

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -67,10 +67,6 @@ add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src" "${CMAKE_BINARY_DIR}/googl
 # COMPILER SETUP
 ######################################################################################################################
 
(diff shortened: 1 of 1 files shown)


hcnelson99:
Hi Wan! Thanks :)

PR #84, opened 2020-08-28
Make clang-format and clang-tidy configs forward-compatible for students running modern clang versions (>8)

hcnelson99:
Make clang-format forward compatible with modern clang (>8) by explicitly setting default options which have changed from clang 8 to clang 10.

Disable clang-tidy options added after clang-8. This prevents students using modern clang versions from getting clang-tidy errors. One alternative way to resolve this issue would be to actually make the modern clang-tidy fixes in the current codebase, but this patch solves the issue without causing any code churn.

Let me know if you guys would rather see the clang-tidy issue solved by actually upstreaming the fixes requested by clang-tidy 10.

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -2,3 +2,7 @@ BasedOnStyle: Google
 DerivePointerAlignment: false
 PointerAlignment: Right
 ColumnLimit: 120
+
+# Default for clang-8, changed in later clangs. Set explicitly for forwards
+# compatibility for students with modern clangs
+IncludeBlocks: Preserve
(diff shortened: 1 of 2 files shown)


lmwnshn:
> Let me know if you guys would rather see the clang-tidy issue solved by actually upstreaming the fixes requested by clang-tidy 10.

@hcnelson99 I'm not officially on the course staff so I won't hit the merge button on that (in case it messes with project development), but that would be nice to have and would probably merge at some point in the future.

hcnelson99:
Yeah I had a feeling that would disrupt in-progress development. Do you need to assign someone on course staff to this PR or can we merge it as-is since CI passes?

PR #87, opened 2020-08-30
P1_2020_rohan

rohanaggarwal7997:


duynguyen-ori75:
Is this possible to retain both versions (clock and LRU)?

After a quick look at both, I think they are pretty similar, so it is quite easy to keep both in the repository.

rohanaggarwal7997:
It is retaining both of them.

Issue #72, opened 2020-08-01
Incorrect comment description for SimpleHashJoinTest


iamkroot:
The SQL query describing the test for SimpleHashJoin Executor is from the [previous test](https://github.com/cmu-db/bustub/blob/f581f10982706f9217957884f97b564d2287f951/test/execution/executor_test.cpp#L201). Not a huge bug, but confused me for a bit.

https://github.com/cmu-db/bustub/blob/f581f10982706f9217957884f97b564d2287f951/test/execution/executor_test.cpp#L257-L258

garrisonhess:
Closing since this comment is no longer present.

PR #88, opened 2020-08-31
fix comments in log_record.h and executor_test.cpp

mrdongmen:
Fix two misleading comments. 

1. The first one is in log_record.h. The NEWPAGE log type has a `page_id` field which was missing in the ASCII graph. I added it.
2. The other one is the comment below SimpleHashJoinTest which is obviously a test on JOIN and not INSERT. I changed the SQL to reflect what the test actually does. 
#72

### DIFF
--- a/src/include/recovery/log_record.h
+++ b/src/include/recovery/log_record.h
@@ -54,9 +54,9 @@ enum class LogRecordType {
  * | HEADER | tuple_rid | tuple_size | old_tuple_data | tuple_size | new_tuple_data |
  *-----------------------------------------------------------------------------------
  * For new page type log record
- *--------------------------
- * | HEADER | prev_page_id |
- *--------------------------
+ *------------------------------------
+ * | HEADER | prev_page_id | page_id |
+ *------------------------------------
  */
 class LogRecord {
   friend class LogManager;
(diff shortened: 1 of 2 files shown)


src/include/recovery/log_record.h
```diff
@@ -54,9 +54,9 @@ enum class LogRecordType {
  * | HEADER | tuple_rid | tuple_size | old_tuple_data | tuple_size | new_tuple_data |
  *-----------------------------------------------------------------------------------
  * For new page type log record
- *--------------------------
- * | HEADER | prev_page_id |
- *--------------------------
+ *------------------------------------
+ * | HEADER | prev_page_id | page_id |
```
duynguyen-ori75, line 58: Actually, I think that the original comment was good.

Considering that the serialization format is used for recovery only, then when you're trying to recover from the last checkpoint, you don't have to worry about the new page id, as the disk manager will do that for you.

There is no point that you request to allocate a new page with a specific page id

mrdongmen: First, this is a description of the log record structure, rather than which part of the structure should be used in your program. It has to be a complete description. I think the error has been introduced in previous versions of this homework. The previous version of the 'NEWPAGE' log record lack the page_id part, it was added later but the comment remains unchanged and mismatches the actual log record structure. 
Second, we are not supposed to discuss how homework details here. Generally, I think when you redo a NEWPAGE, it was already in the disk. You should fetch it from the disk and clean it.

PR #89, opened 2020-09-03
added memory based "disk manager" 

zhou0ricky:
I wasn't sure whether or not it would be wise to touch original disk manager implementation, so I added a separate independent class for the memory based implementation. To make the format better, I could create an interface for both the original disk manager and my implementation to inherit to make testing easier.


rohanaggarwal7997:
Discussed with ricky. We need to do something else. Closing this PR.

PR #99, opened 2020-09-29
Add a CMake safeguard to root CMakeLists.txt.

lmwnshn:
People tend to accidentally run cmake from the wrong folder, which might nuke their source tree and/or cause weird bugs. This forces an out-of-source CMake build by detecting if you are running cmake in a directory that already contains a CMakeLists.txt.

Small snippet extracted from the cmake refactor in NoisePage. You might want this for later projects.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,15 @@ project(BusTub
         LANGUAGES C CXX
         )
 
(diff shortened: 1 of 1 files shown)


portablesounds (approved):
Excellent safeguard for students. Thanks!

PR #100, opened 2020-10-01
Fixed leaf page Lookup signature

githubxxcc:
Fixed leaf page `Lookup` signature.


portablesounds (approved):
Changing to a pointer seems like a better change.

githubxxcc:
> Changing to a pointer seems like a better change.

Oh, this value is to be modified. So that's why.

PR #102, opened 2020-10-04
Add operator= in TableIterator.

ashishnegi:
This prevents `use after free` bugs when objects of TableIterator are assigned to one another.
TableIterator's destructor deletes `Tuple`. So, assignment operator= should create a new Tuple like Copy Constructor.


lmwnshn:
I would be more inclined to delete the copy assignment operator entirely, AFAIK there is not currently a use-case for it.

ashishnegi:
I needed this in my implementation of query executor assignment. So, thought i will add the changes i did.

lmwnshn:
Thanks for the PR, closing as multiple semesters of students have since gone by.

PR #104, opened 2020-10-05
Updated confusing function signatures and add comments

githubxxcc:


githubxxcc:
@portablesounds @YuhongZhang98  Could you please take a look at these?

YuhongZhang98 (approved):

src/include/storage/page/b_plus_tree_leaf_page.h
```diff
@@ -56,17 +56,17 @@ class BPlusTreeLeafPage : public BPlusTreePage {
   int Insert(const KeyType &key, const ValueType &value, const KeyComparator &comparator);
   bool Lookup(const KeyType &key, ValueType *value, const KeyComparator &comparator) const;
   int RemoveAndDeleteRecord(const KeyType &key, const KeyComparator &comparator);
+
   // Split and Merge utility methods
-  void MoveHalfTo(BPlusTreeLeafPage *recipient, BufferPoolManager *buffer_pool_manager /* Unused */);
```
Noeyfan, line 60: Hi,

Seems after this change, We no longer have API parity between leaf_page and internal_page, this means template function like https://github.com/cmu-db/bustub/blob/master/src/storage/index/b_plus_tree.cpp#L130
will require explicit check & cast to compile. This  void the purpose of template function, is it done intentionally?

PR #110, opened 2020-10-13
ignore vagrant configuration directory

ziyi-yan:


lmwnshn:
Do we officially support vagrant? @portablesounds

portablesounds:
Yes we do support Vagrant now via a Vagrantfile in the directory. This seems perfectly reasonable to me.

portablesounds (approved):
Seams reasonable to me as we now officially support Vagrant and wouldn't want those local build files from it to be committed.

Issue #97, opened 2020-09-25
buffer_pool_manager_test.cpp : stcrncmp, strcmp


prateek-agarwal-git:
Line 48, 49: It may be a good idea to use memcpy and memcmp for copying and comparing binary data. String functions do not copy beyond the null character.

prateek-agarwal-git:
Sorry unassigned by mistake.

PR #112, opened 2020-10-15
use strncmp for BinaryDataTest in BufferPoolManagerTest

ziyi-yan:
For comparing all the bytes of pages, we need to use `std::strncmp` with `PAGE_SIZE` as an argument, like what we have done in [here](https://github.com/cmu-db/bustub/blob/449b85fa817005afcb1856d603b1ac480fe2ca90/test/buffer/buffer_pool_manager_test.cpp#L52).


ziyi-yan:
I just found that `memcpy` and `memcmp` are better for this case in order to test binary data saving and retrieving, pointed out by https://github.com/cmu-db/bustub/issues/97. I'm going to close this PR.

PR #113, opened 2020-10-15
Use memcpy and memcmp for BinaryDataTest in BufferPoolManagerTest.

ziyi-yan:
Closes https://github.com/cmu-db/bustub/issues/97

PR #119, opened 2020-11-02
issue#97, using memcmp,memcpy instead of strcmp,strcpy respectively f…

prateek-agarwal-git:
Using memcmp, memcpy instead of strcmp, strcpy respectively for copying and testing random binary data. Resolved issue#97


lmwnshn:
This was fixed in #113 already. Thanks.

PR #130, opened 2020-12-08
Disable brew update in packages.sh

apavlo:
Attempt to avoid the build timeout on Travis for OSX build.

https://youtu.be/5_VtGtg-knI


lmwnshn:
I think Travis is basically dead now.

PR #137, opened 2021-02-04
Fix first_page->Init - Should be INVALID_PAGE_ID

farleyknight:
This is relatively minor, but it looks like the 3rd argument to `TablePage::Init` is the previous page ID. I don't think this triggered any errors because `INVALID_LSN` and `INVALID_PAGE_ID` are both `-1`.


lmwnshn (approved):
LGTM. Good catch.

PR #141, opened 2021-03-16
Use SetUp() and TearDown() to clean up test.db and test.log

greenmoon55:
- Remove test.db and test.log before test run to avoid test
  failures caused by non-empty test.db and test.log from
  other failed tests.
- Add NOLINTNEXTLINE.
- Use SetUp() and TearDown() methods in RecoveryTest 
  and DiskManagerTest.

### DIFF
--- a/test/recovery/recovery_test.cpp
+++ b/test/recovery/recovery_test.cpp
@@ -40,7 +40,6 @@ class RecoveryTest : public ::testing::Test {
     remove("test.db");
     remove("test.log");
   };
-
 };
 
 // NOLINTNEXTLINE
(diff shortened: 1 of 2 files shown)


apavlo (changes requested):
Technically this correct, but you should use gtest `SetUp()` and `TearDown()` methods to clean up these files.

greenmoon55:
> Technically this correct, but you should use gtest `SetUp()` and `TearDown()` methods to clean up these files.

Sure! I can add `SetUp()` and `TearDown()` methods in recovery_test.cpp too.

greenmoon55:
Updated!

apavlo (approved):
LGTM

PR #151, opened 2021-05-13
Changing compliance for zero-length arrays

samwhitlock:
C++ does not support flexible array members. [Here's a good source](http://david.tribble.com/text/cdiffs.htm#C99-fam).

This is only supported by compiler convention.

If you want to abuse C99 things where the C++ compiler plays nice, remove the number altogether i.e. `MappingType array[];`.

Seeing as how this type is always instantiate via a reinterpret cast, this should not be a performance or correctness issue.

The issue with this being undefined behavior is that it may not exist, or may go away if a compiler optimization level deems it to be a candidate for "no-unique-address" behavior.


lmwnshn:
Thanks for the PR! I am not particularly in favor nor in opposition to changing it, since personally I think the compiler extensions / standards discussion is slightly out of scope. I am leaving it open for future TAs to decide. If we do decide to change it, I am in favor of how this PR does it (easier to explain).

garrisonhess (approved):
Thanks for the fix!  I also added some comments clarifying the usage for students.

PR #163, opened 2021-08-30
fix: IndexWriteRecord should keep a record of old tuple

Sorosliu1029:
`IndexWriteRecord` should record old tuple, in order to restore old index key, in case of update operation rollback


garrisonhess (approved):
Looks good, thanks!

PR #176, opened 2021-09-15
Fix initialize instance_index_ in BPI

algebra2k:
In `BufferPoolManagerInstance::ValidatePageId`, the `page_id` mod `num_instances` with `the instance_index` is assert to be equal, but the `instance_index` is not assigned an initial value in **BPI**, which causes its value to always be 0.


mbutrovich (approved):
LGTM

iwanttobepowerful:
@mbutrovich Hi, can reopen fall19's gradescope? Thanks.

apavlo:
@iwanttobepowerful If you are a student at CMU, then please email me. Nobody should be working on fall2019 projects and it was not available outside of CMU.

iwanttobepowerful:
@apavlo OK, thanks for your reply!😊

PR #180, opened 2021-09-26
unnecessary code block removed

thetarby:
```GetFreeSpaceRemaining() < tuple.size_ + SIZE_TUPLE``` check is already done in the beginning of the method and if it is true method returns directly hence when we come to 57th line ```GetFreeSpaceRemaining() < tuple.size_ + SIZE_TUPLE``` is false and it is impossible to get inside of the if block.

### DIFF
--- a/src/storage/page/table_page.cpp
+++ b/src/storage/page/table_page.cpp
@@ -53,11 +53,6 @@ bool TablePage::InsertTuple(const Tuple &tuple, RID *rid, Transaction *txn, Lock
     }
   }
 
-  // If there was no free slot left, and we cannot claim it from the free space, then we give up.
-  if (i == GetTupleCount() && GetFreeSpaceRemaining() < tuple.size_ + SIZE_TUPLE) {
-    return false;
-  }
-
   // Otherwise we claim available free space..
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

**Comment by lmwnshn** [comment]
@garrisonhess points out that we may break early from the loop, which I agree with and missed on first reading -- this might mean that we need to keep the code block.

**Comment by thetarby** [comment]
I am not sure what would breaking early from for loop would change. The if expression is anding ```GetFreeSpaceRemaining() < tuple.size_ + SIZE_TUPLE```, which we know false, with some other thing but that some other thing should not be important because they are and'ed and the whole expression is false regardless of the value of i. 
```cpp
  // Try to find a free slot to reuse.
  uint32_t i;
  for (i = 0; i < GetTupleCount(); i++) {
    // If the slot is empty, i.e. its tuple has size 0,
    if (GetTupleSize(i) == 0) {
      // Then we break out of the loop at index i.
      break;
    }
  }

  //  regardless of value of (i == GetTupleCount()) whole expression is false here
  if (i == GetTupleCount() && GetFreeSpaceRemaining() < tuple.size_ + SIZE_TUPLE) {
     return false;
  }
```

am I missing something @garrisonhess, @lmwnshn ?

PR #185, opened 2021-10-04
Fix overflow in IntComparator

### LINKED ISSUE #184: IntComparator overflow risk
IntComparator could yield incorrect results due to overflow. For example, IntComparator()(INT_MIN, INT_MAX) == 1, but it should be -1 (or any negative number):

https://github.com/cmu-db/bustub/blob/d3cea79bbe49e9daa16c1be081dbc3bd6eb6dbad/src/include/storage/index/int_comparator.h#L17-L24

Also (though slightly off topic), it would be better to mark the member function as const to be consistent with GenericComparator and STL comparators.

Aureliano98:
Fixes #184: Prevents overflow in IntComparator and marks operator() as const

### DIFF
--- a/src/include/storage/index/int_comparator.h
+++ b/src/include/storage/index/int_comparator.h
@@ -20,6 +20,14 @@ namespace bustub {
  */
 class IntComparator {
  public:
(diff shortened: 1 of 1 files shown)


apavlo (approved):
LGTM

PR #189, opened 2021-10-24
Modify clang search path resolution in CMakeLists.

lmwnshn:
Specifically, try to use homebrew binaries if they exist,
before trying to use /usr/local/bin and /usr/bin binaries.

This may fix issues for students who end up using a weird version of clang-tidy.
**This PR needs testing.** I don't have a Mac.

Please check that:

- On a Mac with both AppleClang and LLVM Clang installed,
- This branch picks up a binary provided by LLVM CLang.


lmwnshn:
There are currently no students, punt to future TAs.

PR #194, opened 2021-11-02
Clang Tidy Uninitialized Memory False Positive

turingcompl33t:
Clang Tidy identifies uninitialized memory in some calls to `HashUtil::CombineHashes()` despite the fact that the memory is correctly initialized. This PR zero-initializes the stack-allocated array used within `HashUtil::CombineHashes()` to avoid this false-positive.


mkpjnx (approved):
lgtm

PR #202, opened 2021-11-12
2021-p4: Concurrency Control with deadlock prevention

mkpjnx:


test/concurrency/lock_manager_test.cpp
```diff
@@ -152,119 +152,51 @@ void UpgradeTest() {
...
-      txn_mgr.Abort(txn1);
-    }
-  });
+  // launch the waiter thread
+  std::thread wait_thread{wait_die_task};
 
-  // Sleep for enough time to break cycle
-  std::this_thread::sleep_for(cycle_detection_interval * 2);
+  // wait for txn1 to lock
+  std::this_thread::sleep_for(std::chrono::milliseconds(100));
```
turingcompl33t, line 189: This is a little nitpicky because I'm sure this solution works in practice, but what do you think about using a latch primitive for this instead? The standard library doesn't include a latch until C++20, but we could roll our own with a `std::atomic` very easily, and it removes any potential for nondeterminism.

turingcompl33t (approved):
LGTM. Two comments that we might want to address before release, but both are not absolutely necessary.

PR #204, opened 2021-11-16
Added support for Fedora

josesei:
Changed shell script that installs dependencies to support Fedora. Tested on version 35.

Note: check-clang-tidy fails (probably the criteria updated for the latest version, version 8 is not available through dnf)

### DIFF
--- a/build_support/packages.sh
+++ b/build_support/packages.sh
@@ -8,6 +8,8 @@
 ##
 ## Supported environments:
 ##  * Ubuntu 18.04
+##  * Ubuntu 20.04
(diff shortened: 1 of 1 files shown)


lmwnshn:
Thanks for the PR! Speaking personally, unfortunately I don't think we will be willing/able to maintain official support for multiple platforms, merging this would imply that we did. But we might be able to discuss adding an official "unsupported unofficial hacks that you can try" somewhere, with the caveat that we do not support or debug any issues that happen on those platforms.

PR #206, opened 2021-11-29
Transaction Manager thread safety

mkpjnx:
Added shared locks to prevent potential race conditions on the global transaction map - this has been causing spurious failures.


lmwnshn (approved):
Rubber stamp.

PR #208, opened 2021-12-11
Support Apple m1 chip

kriti-sc:
This PR add support for building BusTub on Apple M1 chip:
1. Upgrades LLVM version to 12 from 8. LLVM 8 does not have Apple M1 support.
2. Specify Apple M1 specific flags for compilation.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -83,7 +83,7 @@ add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src" "${CMAKE_BINARY_DIR}/googl
 ######################################################################################################################
 
 # Compiler flags.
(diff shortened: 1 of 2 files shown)


StrongerXi:
I don't think `-mcpu=apple-a14` is _adding_ support for building on Apple M1 chip. Rather it seems to be restricting build on M1 chip only.

I ran into the same issue, and for your reference, simply removing `-march=native` enabled me to build the project on M1 chip MBA.

The most general solution is to use Docker though, I'd say.

apavlo:
This breaks the current build on Linux x86 and we cannot merge.

Our cmake setup is based on Apache Arrow. I would look to see how they fix it or wait until fall 2022 when we will fix it:

https://github.com/apache/arrow/tree/master/cpp

PR #218, opened 2022-02-05
fix log bug in directory page Verify func

zdszero:
just as the diff shows, the logged variable should be curr_count


apavlo (approved):
LGTM

PR #223, opened 2022-02-25
feat: add the size of the display b-tree node for the ToGraph method

algebra2k:
Let `ToGraph` display the size of b-plus-tree nodes for better debugging.


apavlo (commented):
LGTM

PR #224, opened 2022-03-10
fix: executor_test assertion for SimpleUpdateTest

pidb:
https://github.com/cmu-db/bustub/blob/9b0da5ebe0985de726ee299b7bec60dc56f1086a/test/execution/executor_test.cpp#L314-L315

The UpdateExecutor really shouldn't modify the result set, but I think the result set of size is TEST3_SIZE.

### DIFF
--- a/test/execution/executor_test.cpp
+++ b/test/execution/executor_test.cpp
@@ -312,7 +312,7 @@ TEST_F(ExecutorTest, DISABLED_SimpleUpdateTest) {
   GetExecutionEngine()->Execute(update_plan.get(), &result_set, GetTxn(), GetExecutorContext());
 
   // UpdateExecutor should not modify the result set
-  ASSERT_EQ(result_set.size(), 0);
+  ASSERT_EQ(result_set.size(), TEST3_SIZE);
   result_set.clear();
 
   // Execute another sequential scan; no tuples should be present in the table


jkosh44:
@7bytes Thanks for the PR! Can you please provide an explanation as to why the size would be `TEST3_SIZE`? I haven't looked at this in a little while, but to me it looks like the result set is cleared on line 309 and since the update executor shouldn't modify the result set, it should remain empty in the assertion.

7bytes:
> @7bytes Thanks for the PR! Can you please provide an explanation as to why the size would be `TEST3_SIZE`? I haven't looked at this in a little while, but to me it looks like the result set is cleared on line 309 and since the update executor shouldn't modify the result set, it should remain empty in the assertion.

@jkosh44 Thank you reply for this pr. Yes, as you said, the update executor should not modify the result set, I think the main reasons why it might need to be modified are:

Here the `result_set` is populated after scan_plan is executed, and

https://github.com/cmu-db/bustub/blob/9b0da5ebe0985de726ee299b7bec60dc56f1086a/test/execution/executor_test.cpp#L295-L301

then `update_plan` and with it child `seq_plan`  execute update

https://github.com/cmu-db/bustub/blob/9b0da5ebe0985de726ee299b7bec60dc56f1086a/test/execution/executor_test.cpp#L309-L315

Although line 309 `reset_set.clear` is executed here, but calling `child_executor.Next()` is to iterate over `TEST3_SIZE` tuples, according to the behavior here,

https://github.com/cmu-db/bustub/blob/9b0da5ebe0985de726ee299b7bec60dc56f1086a/src/include/execution/execution_engine.h#L59-L66

It populates `result_set` unless it returns false, but returning false should be an error because `child_executor.Next()` doesn't go to the end.

So the `result_set size` should still be `TEST3_SIZE`, but the tuple in the result set should not be overwritten by update executor, I think so :)

turingcompl33t:
Thanks for the interest in the project @7bytes. The test logic is correct; you should modify the body of `ExecutionEngine::Execute()` to ensure that the result set is not updated to match the specification of the test.

lmwnshn:
I haven't looked at it personally, but I think the consensus is that the test is valid.

PR #242, opened 2022-05-24
Add empty check in BPlusTree Print and Draw funcs

IcePigZDB:
If it's an empty tree, we can give some DEBUG hints and return.

### DIFF
--- a/src/include/storage/index/b_plus_tree.h
+++ b/src/include/storage/index/b_plus_tree.h
@@ -62,7 +62,7 @@ class BPlusTree {
 
   void Print(BufferPoolManager *bpm) {
     if (IsEmpty()) {
-      LOG_DEBUG("Print an empty tree");
+      LOG_WARN("Print an empty tree");
       return;
     }
     ToString(reinterpret_cast<BPlusTreePage *>(bpm->FetchPage(root_page_id_)->GetData()), bpm);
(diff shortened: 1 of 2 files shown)


src/include/storage/index/b_plus_tree.h
```diff
@@ -60,10 +61,18 @@ class BPlusTree {
...
+    if (IsEmpty()) {
+      LOG_DEBUG("Print an empty tree");
+      return;
+    }
     ToString(reinterpret_cast<BPlusTreePage *>(bpm->FetchPage(root_page_id_)->GetData()), bpm);
   }
 
   void Draw(BufferPoolManager *bpm, const std::string &outf) {
+    if (IsEmpty()) {
+      LOG_DEBUG("Draw an empty tree");
```
apavlo, line 73: Change this to a warning instead of a debug message

apavlo (changes requested):

src/include/storage/index/b_plus_tree.h
```diff
@@ -14,6 +14,7 @@
 #include <string>
 #include <vector>
 
+#include "common/logger.h"
```
lmwnshn, line 17: I'm not very enthusiastic about including logger in a header file because that probably means it'll get included everywhere. But I guess reducing compilation time can be a future problem.

lmwnshn (approved):
Requested changes were made, lgtm

lmwnshn:
Looks like I don't have the power to override @apavlo requesting changes, so I can't merge.

src/include/storage/index/b_plus_tree.h
IcePigZDB: Hi, I move the implemention and include of logger.h to b_plus_tree.cpp, ptal.

lmwnshn (approved):
Thanks for making the change!

PR #243, opened 2022-05-25
Toolchain refresh - Support Ubuntu 22.04 and latest macOS (including M1)

garrisonhess:
Problem: Toolchain is old enough that neither Ubuntu 22.04 nor the M1 mac work out-of-the-box.  This PR fixes that problem.  Here are some relevant issues: #231 #153 #173.

Solution: This PR updates the toolchain to use LLVM/Clang version 12 and G++ version 12.  These are all available via brew and apt, so they are suitable for the environments we aim to support.  I have tested this on Ubuntu 22.04 with x86-64 and macOS Monterey (12) with the M1 (ARM) processor (current MacBook pro).

Change list and comments: 
- Update LLVM/Clang from 8 to 12
- Update G++ from 7 to 12 (mostly because it's readily available on apt/brew, whereas 7 is not).
- Fix a syntax bug in packages.sh.
- Disable clang-tidy-12 checks `readability-function-cognitive-complexity, bugprone-suspicious-include, bugprone-narrowing-conversions, bugprone-reserved-identifier`.  I have a separate branch that fixes all these lints, but it is over 1500 lines so I have omitted it from this PR.
- Remove `-march=native` from clang flags because it requires Clang version 15+ to work with macOS M1 processors. 
LLVM issue reference: 
https://discourse.llvm.org/t/why-does-march-native-not-work-on-apple-m1/2733
LLVM fix reference: https://www.phoronix.com/scan.php?page=news_item&px=LLVM-Clang-march-native-M1
- Change all `#!/usr/env python` shell directives to `#!/usr/env python3` because this is the correct way to invoke python3 on the latest Ubuntu and macOS versions.  To elaborate, Ubuntu does not come with `python` but it does come with `python3`.  Similarly, if one wants`python3` on vanilla macOS, then `python` won't work.  This was a common issue in OH last year.

PR #244, opened 2022-05-25
[WIP] CI - Cover target environments fully.

garrisonhess:
Problem: The current Github Actions configuration is missing Ubuntu 18.04 and 22.04, and also macOS 12.

Solution: Add them in.

Note: This will not work without the changes from #243, so it is essentially pending on those changes.


garrisonhess:
Moved to #245

PR #246, opened 2022-05-29
Apply clang-tidy modernize-use-trailing-return-type

garrisonhess:
This PR converts all function signatures to use a trailing return type.  Trailing return types were introduced in C++11 and are a new check included in the upgrade from clang-8 to clang-12.


apavlo (approved):
LGTM

PR #252, opened 2022-06-24
fix-executor-test

Rachelint:
As I think, in SimpleUpdateTest，we should remove the check and pass nullptr as result_set when executing update plan.
Because executor->Next() will return true before all tuples are  processed and tuple will be pushed into result_set continuously In GetExecutionEngine()->Execute() when passed result_set is not null, result_set's size will not be zero after executing and we can't pass this test. And we don't care the result_set when updating actually. 
So I think passing nullptr and not checking may be reasonable.


Rachelint:
We can update all tuples in one Next() calling which return false and pass the orginal test，but it doesn't quite fit the volcanic model. Is there anything wrong with my understanding？

Issue #150, opened 2021-04-13
why open mode has both set trunc and app when create a file in Diskmanager


tang-hi:
I notice when file didn't exist, DiskManager will create file in disk_manager.cpp ,line 46 
`log_io_.open(log_name_, std::ios::binary | std::ios::trunc | std::ios::app | std::ios::out);`
But when I use it to create file, it doesn't work because C++ didn't allow to both set trunc and app when open the file.
[https://stackoverflow.com/questions/26900392/why-does-fstream-open-fail-if-the-mode-has-both-trunc-and-app-set](so)
[https://www.cplusplus.com/reference/fstream/fstream/open/](url)
![Screenshot from 2021-04-13 21-35-00](https://user-images.githubusercontent.com/72755185/114562162-d65c3680-9ca0-11eb-8e8b-f6adbdcc2ce6.png)

I think  it may be should use `log_io_.open(log_name_, std::ios::binary | std::ios::trunc | std::ios::out);` instead.

skyzh:
I believe it's a bug. Indeed you'll also need `std::ios::in` in addition to `std::ios::binary | std::ios::trunc | std::ios::out` to open the file. I've proposed a PR to fix it. You may take a look if you're interested :)

tang-hi:
> I believe it's a bug. Indeed you'll also need `std::ios::in` in addition to `std::ios::binary | std::ios::trunc | std::ios::out` to open the file. I've proposed a PR to fix it. You may take a look if you're interested :)

thanks,LGTM

PR #256, opened 2022-07-27
fix(disk_manager): remove redundant open file

skyzh:

close https://github.com/cmu-db/bustub/issues/150

```
log_io_.open(log_name_, std::ios::binary | std::ios::trunc | std::ios::app | std::ios::out);
```

This line will always fail and doesn't take effect at all. We can directly create (or truncate) a file if we can't open it.


garrisonhess:
Does this work with the refsols?

skyzh:
Not tested yet, but I assume the behavior won’t change, because the removed line didn’t take any effect.

garrisonhess (approved):
LGTM!

PR #259, opened 2022-08-04
Disable 2 clang rules for refsol compliance

garrisonhess:
This PR brings the reference solution into compliance with newer versions of clang-tidy.  Compliance with these rules would only involve changing test names and the random number generation API, so it can be done reasonably quickly in the future.


lmwnshn (approved):
LGTM

Issue #260, opened 2022-08-07
executor_test.cpp SimpleGroupbyAggregation


sg20180546:
hello . I glad you to open this bustub open source educational project to public.
It helps me to understand database internal a lot

during solving the executor_test.cpp,
Test SimpleGroupbyAggregation seems little weird to me.
That is, the test demands `count(col_a), col_b, sum(col_c) FROM test_1 Group By col_b HAVING count(col_a) > 100`
but the output schema init as `agg_schema = MakeOutputSchema({{"countA", count_a}, {"colB", groupby_b}});`

I think that correct output schema shoud initialized as
`agg_schema = MakeOutputSchema({{"countA", count_a}, {"colB", groupby_b}, {"sumC", sum_c} });`

Thanks.

skyzh:
That’s true, I think there’s something wrong with that. I’ll propose a fix later.

PR #261, opened 2022-08-07
fix(executor): add sum column for test

skyzh:

close https://github.com/cmu-db/bustub/issues/260

tested with sample solution, which works fine.

By the way, it seems that all our execution tests' test data are randomly generated (table_generator.cpp), which might make things hard when someone want to debug their code and see what's wrong. I will look into the possibility of having some tables of fixed value in our tests in the future if possible.

### DIFF
--- a/test/execution/executor_test.cpp
+++ b/test/execution/executor_test.cpp
@@ -581,7 +581,8 @@ TEST_F(ExecutorTest, DISABLED_SimpleGroupByAggregation) {
   for (const auto &tuple : result_set) {
     // Should have count_a > 100
     ASSERT_GT(tuple.GetValue(agg_schema, agg_schema->GetColIdx("countA")).GetAs<int32_t>(), 100);
-    // Should have sum_c >= 0
+    // Should have sum_c >= 0. Data for test_1 table is randomly generated, where colC is uniformly distributed from
+    // 0 to 9999. So we can only ensure sumC column exists by checking if it's >= 0 here.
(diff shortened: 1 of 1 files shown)


skyzh:
> ```
> // Should have sum_c >= 0
> ```

... because table test_1's colC has uniform distribution of 0-9999. So we can only assert >=0 here. The check is somehow meaningless, but at least we can check if the column is present or not.

lmwnshn:
On randomly generated test data: it is OK if you make the student-facing tests deterministic and easy to debug, but the private grading tests should remain effectively random and unguessable by students. This is to prevent people from hard-coding answers by trying to leak test information from autograder output.

skyzh:
Not sure whether sum_c aggregator is using column A or C. I'll look into the code and ensure this fix is correct before merging.

lmwnshn:
> ... because table test_1's colC has uniform distribution of 0-9999. So we can only assert >=0 here. The check is somehow meaningless, but at least we can check if the column is present or not.

I would add this to the test comments directly so that students do not puzzle over what the test is trying to do.

skyzh:
> Not sure whether sum_c aggregator is using column A or C. I'll look into the code and ensure this fix is correct before merging.

It should be using colC now. Also added comments about the sumC >= 0 check.

garrisonhess:
Is this confirmed to work with the refsols?  It looks good to me, I just want to verify before approval.

skyzh:
Yes it works for refsol.

garrisonhess (approved):
Good addition, thanks!

PR #269, opened 2022-08-19
refactor(build): use separate CMakeLists for each component

skyzh:

This is part of https://github.com/garrisonhess/bustub-public/pull/20, which will eventually make us easier to introduce DuckDB's parser. <del>Still working on integrating it with our ref solutions.</del>

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,10 +78,6 @@ execute_process(COMMAND "${CMAKE_COMMAND}" --build .
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)  # don't override our compiler/linker options when building gtest
 add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src" "${CMAKE_BINARY_DIR}/googletest-build")
 
-function(disable_target_warnings NAME)
-        target_compile_options(${NAME} PRIVATE "-w")
-endfunction()
-
 ######################################################################################################################
 # COMPILER SETUP
 ######################################################################################################################
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,6 +78,10 @@ execute_process(COMMAND "${CMAKE_COMMAND}" --build .
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)  # don't override our compiler/linker options when building gtest
 add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src" "${CMAKE_BINARY_DIR}/googletest-build")
 
+function(disable_target_warnings NAME)
+        target_compile_options(${NAME} PRIVATE "-w")
+endfunction()
+
 ######################################################################################################################
 # COMPILER SETUP
 ######################################################################################################################
@@ -101,6 +105,7 @@ set(CMAKE_ARCHIVE_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_LIBRARY_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/lib)
 set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
+
 # Includes.
 set(BUSTUB_SRC_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/src/include)
 set(BUSTUB_TEST_INCLUDE_DIR ${PROJECT_SOURCE_DIR}/test/include)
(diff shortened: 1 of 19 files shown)


CMakeLists.txt
```diff
@@ -78,6 +78,10 @@ execute_process(COMMAND "${CMAKE_COMMAND}" --build .
 set(gtest_force_shared_crt ON CACHE BOOL "" FORCE)  # don't override our compiler/linker options when building gtest
 add_subdirectory("${CMAKE_BINARY_DIR}/googletest-src" "${CMAKE_BINARY_DIR}/googletest-build")
 
+function(disable_target_warnings NAME)
+        target_compile_options(${NAME} PRIVATE "-w")
```
garrisonhess, line 82: It looks like the only use of this is commented out.  Maybe we can remove this?

src/CMakeLists.txt
```diff
@@ -1,13 +1,38 @@
-# This is hacky :(
-file(GLOB_RECURSE bustub_sources ${PROJECT_SOURCE_DIR}/src/*/*.cpp ${PROJECT_SOURCE_DIR}/src/*/*/*.cpp)
```
garrisonhess, line 2: I was worried that removing this line would cause any new files in the refsol not to be built properly.  Does this PR work with the refsols?

src/container/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+add_subdirectory(hash)
```
garrisonhess, line 1: This is one example of places where I felt that we might be able to organize things better.  But it's probably not worth changing now.

test/CMakeLists.txt
```diff
@@ -45,7 +46,8 @@ foreach (bustub_test_source ${BUSTUB_TEST_SOURCES})
     add_dependencies(build-tests ${bustub_test_name})
     add_dependencies(check-tests ${bustub_test_name})
 
-    target_link_libraries(${bustub_test_name} bustub_shared gtest gmock_main)
+    # target_link_libraries(${bustub_test_name} bustub bustub_buffer bustub_catalog bustub_common_util bustub_common bustub_concurrency bustub_execution bustub_parser bustub_recovery bustub_type bustub_murmur3 bustub_pg_query bustub_sqlite3 gtest gmock_main)
```
garrisonhess, line 49: Should we remove this dead code?

third_party/murmur3/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+cmake_minimum_required(VERSION 3.0)
+# project(bustub_murmur3 CXX C)
+
+# set(CMAKE_CXX_VISIBILITY_PRESET hidden)
+
+add_library(bustub_murmur3 SHARED MurmurHash3.cpp)
+
+# install(TARGETS bustub_murmur3
```
garrisonhess, line 8: More dead code here.  We will want to resolve this before merging anything.  I think checking things with the reference solutions will help inform what is needed.  Running things on bustub public without real implementation likely means that the murmur3 library is never being used (since there are no real backend implementations).

third_party/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+# don't export any symbols from the third party stuff
+# set(CMAKE_C_VISIBILITY_PRESET hidden)
```
garrisonhess, line 2: More dead code here.  Let's resolve this before merging.

garrisonhess (dismissed):
I agree that the CMake refactoring should be done first.  Good work on splitting the query processing PR into smaller chunks!

It looks like there is still some dead / commented out stuff in some of the CMakeLists.txt files.  It would be good to refine this into only the required changes.

Also, have you confirmed this to work with the refsols?

Lastly, it may be worth investigating ways to clean up the project layout.  I don't think it's necessary for this PR, as it will only create more work.  But it may help make the project cleaner and easier to extend.  I remember finding DuckDB's project layout very intuitive, and I think moving in that direction could be a good thing.

src/CMakeLists.txt
skyzh: There’s another PR in bustub private. We will also update CMakeLists if built with refsol.

test/CMakeLists.txt
skyzh: +1 will do

skyzh:
Updated and confirmed that it works with refsol on Ubuntu 22.04.

src/container/CMakeLists.txt
skyzh: I'll look into it later :)

lmwnshn (approved):
From an extremely quick skim, and without running any code or even checking out this branch, I see no issue with this organization. DuckDB has a very nice wrapper around add_library that can do unity builds, IIRC -- may want to yoink that at some point for significantly better compile times.

skyzh:
Sounds good to me. I'll see if we can integrate with CMake's unity builds in the future.

PR #273, opened 2022-08-21
feat: add shell and introduce trivial binder 

skyzh:
follow-up on https://github.com/cmu-db/bustub/pull/269. need rebase to main before reviewing

This is part of https://github.com/garrisonhess/bustub-public/pull/20. In this PR, we:

* Vendor libpg_query from DuckDB.
* Vendor fmt.
* Introduce a very simple binder based on https://github.com/garrisonhess/bustub-public/pull/20. The parser in Garrison's PR is indeed a binder. We can directly use Postgres's parser implementation without introducing DuckDB's parser directory.
* Add a very simple shell (instead of introducing sqllite_wrapper) to print binder result. Will make it a real SQL shell in future PRs.
* Add a binder framework, bind table name in select statements.


skyzh:
In this PR, we introduce a binder framework (though not complete).

image

Currently, it can correctly pick up the table name from the select statement. We can extend it to parse / bind all queries we want to support in the future.

skyzh:
... and please rebase and merge for this PR to separate the commits of vendor and actual impl.

lmwnshn (approved):
We are merging this before the class starts but practically speaking I haven't reviewed this. Just going to rubber stamp for now.

Issue #272, opened 2022-08-21
Discussion: check clang-tidy on changed files


skyzh:
clang-tidy is so slow. I propose to do clang-tidy checks when...

* submitting PR -- only check on changed files
* merged to master -- check all files
* on student submission -- check submitted files

In this way, we can greatly improve development experience.

PR #274, opened 2022-08-22
chore: run clang-tidy on git diff

skyzh:

close https://github.com/cmu-db/bustub/issues/272

This PR adds a new Makefile command `check-clang-tidy-diff`, which runs clang-tidy only on files diff to origin/master.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,4 +176,10 @@ add_custom_target(check-clang-tidy
         -clang-tidy-binary ${CLANG_TIDY_BIN}                              # using our clang-tidy binary
         -p ${CMAKE_BINARY_DIR}                                            # using cmake's generated compile commands
         )
+add_custom_target(check-clang-tidy-diff
+        ${BUSTUB_BUILD_SUPPORT_DIR}/run_clang_tidy.py                     # run LLVM's clang-tidy script
+        -clang-tidy-binary ${CLANG_TIDY_BIN}                              # using our clang-tidy binary
+        -p ${CMAKE_BINARY_DIR}                                            # using cmake's generated compile commands
+        -only-diff                                                        # only check diff files to master
+        )
 add_dependencies(check-clang-tidy gtest bustub)                    # needs gtest headers, compile_commands.json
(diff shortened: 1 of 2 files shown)


CMakeLists.txt
```diff
@@ -173,4 +173,10 @@ add_custom_target(check-clang-tidy
         -clang-tidy-binary ${CLANG_TIDY_BIN}                              # using our clang-tidy binary
         -p ${CMAKE_BINARY_DIR}                                            # using cmake's generated compile commands
         )
+add_custom_target(check-clang-tidy-diff
+        ${BUSTUB_BUILD_SUPPORT_DIR}/run_clang_tidy.py                     # run LLVM's clang-tidy script
+        -clang-tidy-binary ${CLANG_TIDY_BIN}                              # using our clang-tidy binary
+        -p ${CMAKE_BINARY_DIR}                                            # using cmake's generated compile commands
+        -only-diff                                                        # Only check diff files to master
```
lmwnshn, line 180:
```suggestion
        -only-diff                                                        # only check diff files to master
```

Consistent capitalization.

build_support/run_clang_tidy.py
```diff
@@ -267,6 +275,17 @@ def main():
     files = [make_absolute(entry['file'], entry['directory'])
              for entry in database]
 
+    if args.only_diff:
```
lmwnshn, line 278: Add a quick short comment before this block to note:

- We added this code
- Why we added this code
- If you ever expect someone to modify this, or if this is porcelain (I don't remember what git commands are porcelain vs plumbing), a quick description of what this code does

1 line each or less is fine, nothing fancy. This is just to make sure that if people replace this with a new run_clang_tidy script in the future, they won't wonder why only_diff is gone.

lmwnshn (changes requested):
LGTM pending minor nits, please also modify the run_clang_tidy.py header to note your changes.

skyzh:
Updated and resolved comments. Indeed the file header mentions that there's something called [clang-tidy-diff.py](https://github.com/llvm-mirror/clang-tools-extra/blob/master/clang-tidy/tool/clang-tidy-diff.py). We can also look into possibility with integrating with that script.

lmwnshn (approved):
LGTM

Issue #190, opened 2021-10-24
Prevent usage of AppleClang.


lmwnshn:
I remember nothing but issues with AppleClang.

We should add a check to https://github.com/cmu-db/bustub/blob/master/CMakeLists.txt that just refuses to build with AppleClang entirely, since we ask them to install LLVM's Clang on OSX anyway.

PR #282, opened 2022-08-26
chore(build): warn on non clang-12 compilers

skyzh:

close https://github.com/cmu-db/bustub/issues/190

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,7 +5,7 @@ set(CMAKE_CXX_STANDARD 17)              # Compile as C++17
 set(CMAKE_CXX_STANDARD_REQUIRED ON)     # Require C++17 support
 
 project(BusTub
-        VERSION 2022.1
+        VERSION 2022.8
         DESCRIPTION "The BusTub Relational Database Management System (Educational) @ https://github.com/cmu-db/bustub"
         LANGUAGES C CXX
         )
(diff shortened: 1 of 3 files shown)


skyzh:
image

... and it turned out we're not using clang-12 on "Ubuntu 22.04 Clang" CI builds! I'll also fix this...

skyzh:
Successful configuring cmake will look like:

image

CMakeLists.txt
```diff
@@ -1,57 +1,72 @@
 cmake_minimum_required(VERSION 3.10)
-set(CMAKE_EXPORT_COMPILE_COMMANDS ON)   # For clang-tidy.
+set(CMAKE_EXPORT_COMPILE_COMMANDS ON) # For clang-tidy.
 set(BUILD_SHARED_LIBS ON)
-set(CMAKE_CXX_STANDARD 17)              # Compile as C++17
-set(CMAKE_CXX_STANDARD_REQUIRED ON)     # Require C++17 support
+set(CMAKE_CXX_STANDARD 17) # Compile as C++17
```
lmwnshn, line 4:
```suggestion
set(CMAKE_CXX_STANDARD 17) # Compile as C++17.
```

CMakeLists.txt
```diff
@@ -1,57 +1,72 @@
...
+set(CMAKE_CXX_STANDARD_REQUIRED ON) # Require C++17 support
```
lmwnshn, line 5:
```suggestion
set(CMAKE_CXX_STANDARD_REQUIRED ON) # Require C++17 support.
```

CMakeLists.txt
```diff
@@ -1,57 +1,72 @@
...
+
+if("${CLANG_FORMAT_BIN}" STREQUAL "CLANG_FORMAT_BIN-NOTFOUND")
     message(WARNING "BusTub/main couldn't find clang-format.")
 else()
     message(STATUS "BusTub/main found clang-format at ${CLANG_FORMAT_BIN}")
 endif()
 
 # clang-tidy
-if (NOT DEFINED CLANG_TIDY_BIN)
+if(NOT DEFINED CLANG_TIDY_BIN)
```
lmwnshn, line 62: Are we using a different formatter?

lmwnshn (changes requested):
Mostly nits, overall LGTM

CMakeLists.txt
skyzh: Yep, I'll switch back to the CLion one.

skyzh:
I think I've made a wrong choice to force students to use clang-12 on macOS.

https://github.com/cmu-db/bustub/pull/285

ASAN with leak sanitizer will always report false positives with homebrew-installed LLVM 12. But it works with AppleClang.

```
3: Direct leak of 32 byte(s) in 1 object(s) allocated from:
3:     #0 0x10148cb6c in wrap_calloc+0x94 (libclang_rt.asan_osx_dynamic.dylib:arm64+0x3cb6c)
3:     #1 0x1a1ed71e4 in cache_t::insert(objc_selector*, void (*)(), objc_object*)+0x148 (libobjc.A.dylib:arm64+0x201e4)
3:     #2 0xd1090001a1ebede4  (<unknown module>)
```

I guess it's because the dynamically linked artifacts in system (like sanitizer libraries) is built with AppleClang. So maybe it would be a better choice to let students use AppleClang on macOS? (But recommend them to use homebrew's clang-tidy)

skyzh:
... and AppleClang reports

```
20: ==71198==AddressSanitizer: detect_leaks is not supported on this platform.
```

It's just unsupported.

skyzh:
ready for review.

lmwnshn (approved):
Approved for the warning, mostly skimmed everything else. What's our final guidance for ASAN on macOS? Can we stabilize this before p1?

skyzh:
macOS only has ASAN without LSAN. Linux has ASAN with LSAN built-in. Students can use AppleClang or clang-12, I don't see any issues with them for now with refsol.

The only difference is that, bustub built with clang-12 on macOS will produce strange output when LSAN is forcibly enabled by developers, while AppleClang won't allow developers to enable LSAN. As the default behavior on macOS is to make LSAN disabled with both of the compilers, I'd be happy with any compiler.

Also our CI is currently tested against AppleClang instead of clang-12 on macOS.

skyzh:
I think this can be stabilized. Students are "recommended" to use clang-12 on macOS, but they're also fine to use AppleClang. LSAN is disabled by default on macOS, no matter which compiler is used.

PR #280, opened 2022-08-25
chore(build): deprecate valgrind in builds

skyzh:
Signed-off-by: Alex Chi <iskyzh@gmail.com>

Still verifying with refsol.

close https://github.com/cmu-db/bustub/issues/277
close https://github.com/cmu-db/bustub/issues/199
close https://github.com/cmu-db/bustub/issues/135

PR #285, opened 2022-08-26
chore(build): enable leak sanitizer in CI

skyzh:

Follow up to deprecating valgrind https://github.com/cmu-db/bustub/pull/280, we should enable leak sanitizer (along with memory sanitizer) in ASAN.


skyzh:
... note that leak sanitizer will produce strange result on macOS

skyzh:
Using `ASAN_OPTIONS=help=1` we can find that leak detectors are by-default enabled on Linux clang / gcc, and by-default disabled on macOS. So we don't need to provide an extra args here. Will close this PR after I confirm everything.

PR #287, opened 2022-08-27
feat: add 2022 fall p0 trie

amstqq:
Pushed boilerplate p0 code and subset of tests.

PR #289, opened 2022-08-27
feat(build): add check-clang-tidy-p0 for p0

skyzh:
Add a new Makefile entry for checking 2022 Fall's p0. Will check only one file so it's significantly faster. Will use this Makefile entry in gradescope. <del>Can merge before https://github.com/cmu-db/bustub/pull/287.</del>


skyzh:
Also fixes the clang-tidy bug that causes a dead loop.

skyzh:
depends on https://github.com/cmu-db/bustub/pull/287

amstqq (approved):

PR #295, opened 2022-08-29
feat(binder): bind simple scan, agg, and expressions

skyzh:

Now we can bind simple queries like `select 1`, `select sum(x) from table group by y having sum(x) > 10`.


lmwnshn:
Converted to draft, please unconvert when ready for review.

skyzh:
Nearly ready to review, adding comments and docs.

skyzh:
Ready for review. Feel free to ask for more docs if necessary.

CMakeLists.txt
```diff
@@ -1,6 +1,6 @@
 cmake_minimum_required(VERSION 3.10)
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)   # For clang-tidy.
-set(BUILD_SHARED_LIBS ON)
+set(BUILD_SHARED_LIBS OFF)
```
skyzh, line 3: libfmt's linker behavior is controlled by this option, so we have to set it as OFF if we want to link everything statically.

CMakeLists.txt
```diff
@@ -173,7 +174,7 @@ add_custom_target(check-lint echo '${BUSTUB_LINT_FILES}' | xargs -n12 -P8
         ${CPPLINT_BIN}
         --verbose=2 --quiet
         --linelength=120
-        --filter=-legal/copyright,-build/header_guard
+        --filter=-legal/copyright,-build/header_guard,-runtime/references
```
skyzh, line 177: The lint is annoying and strange on templates `const auto &T`.

src/include/common/config.h
```diff
@@ -27,14 +27,16 @@ extern std::atomic<bool> enable_logging;
...
 extern std::chrono::duration<int64_t> log_timeout;
 
+static constexpr const uint64_t INVALID_INDEX = -1;
 static constexpr int INVALID_PAGE_ID = -1;                                    // invalid page id
 static constexpr int INVALID_TXN_ID = -1;                                     // invalid transaction id
 static constexpr int INVALID_LSN = -1;                                        // invalid log sequence number
 static constexpr int HEADER_PAGE_ID = 0;                                      // the header page id
 static constexpr int PAGE_SIZE = 4096;                                        // size of a data page in byte
-static constexpr int BUFFER_POOL_SIZE = 10;                                   // size of buffer pool
+static constexpr int BUFFER_POOL_SIZE = 64;                                   // size of buffer pool
```
skyzh, line 36: changed because tables cannot be generated with the default config (reporting cannot create new page). need investigation later.

CMakeLists.txt
lmwnshn: As a general suggestion, IMO comments like this belong next to the code (where people will see them) and not on the github PR (where people won't)

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "recovery/log_manager.h"
+
+namespace bustub {
+
+auto BustubInstance::GetExecutionContext(Transaction *txn) -> std::unique_ptr<ExecutorContext> {
+  return std::make_unique<ExecutorContext>(txn, catalog_, buffer_pool_manager_, transaction_manager_, lock_manager_);
+}
+
+BustubInstance::BustubInstance(const std::string &db_file_name) {
+  enable_logging = false;
```
lmwnshn, line 14: Please add a comment describing why

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+  // storage related
```
lmwnshn, line 16:
```suggestion
  // Storage related.
```

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  disk_manager_ = new DiskManager(db_file_name);
+
+  // log related
```
lmwnshn, line 19:
```suggestion
  // Log related.
```

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  log_manager_ = new LogManager(disk_manager_);
+
+  buffer_pool_manager_ = new BufferPoolManagerInstance(BUFFER_POOL_SIZE, disk_manager_, log_manager_);
+
+  // txn related
```
lmwnshn, line 24:
```suggestion
  // Transaction (txn) related.
```

I normally wouldn't bother spelling txn out, but it is plausible that students will read this code.

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  lock_manager_ = new LockManager();
+  transaction_manager_ = new TransactionManager(lock_manager_, log_manager_);
+
+  // checkpoints
```
lmwnshn, line 28:
```suggestion
  // Checkpoint related.
```

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  checkpoint_manager_ = new CheckpointManager(transaction_manager_, log_manager_, buffer_pool_manager_);
+
+  // catalog
```
lmwnshn, line 31:
```suggestion
  // Catalog.
```

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  catalog_ = new Catalog(buffer_pool_manager_, nullptr, nullptr);
+}
+
+auto BustubInstance::ExecuteSql(const std::string &sql) -> std::vector<std::string> {
+  if (!sql.empty() && sql[0] == '\\') {
+    // internal commands
```
lmwnshn, line 37:
```suggestion
    // Internal meta-commands, like in `psql`.
```

Since we're already borrowing PostgreSQL terminology, we might as well keep doing that.

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+    if (sql == "\\dt") {
+      std::string result;
```
lmwnshn, line 39: Something something stringstream. eh... because students might read this, I think we should do it.

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
...
+  bustub::Binder binder;
+  binder.ParseAndBindQuery(sql, *catalog_);
+  std::vector<std::string> result = {};
+  for (auto &&statement : binder.statements_) {
+    result.push_back(statement->ToString());
+  }
+  return result;
+}
+
+void BustubInstance::GenerateTestTable() {
```
lmwnshn, line 59: We could move this to a test fixture, right?

src/common/bustub_instance.cpp
```diff
@@ -0,0 +1,80 @@
+#include "common/bustub_instance.h"
+#include <fmt/format.h>
```
lmwnshn, line 2: Quotes now that we vendor?

src/binder/statement/select_statement.cpp
```diff
@@ -1,37 +1,325 @@
 #include "binder/statement/select_statement.h"
```
lmwnshn, line 1: This is the last file I am reviewing and I am starting to become sleepy again, so it'll be very lightly reviewed..

src/binder/statement/select_statement.cpp
```diff
@@ -1,37 +1,325 @@
...
-    auto node = reinterpret_cast<duckdb_libpgquery::PGNode *>(c->data.ptr_value);
-    switch (node->type) {
-      case duckdb_libpgquery::T_PGRangeVar: {
-        if (found) {
-          throw Exception("shouldnt have multiple tables");
+  // bind select list
+  if (pg_stmt->targetList != nullptr) {
+    BindSelectList(pg_stmt->targetList);
+  } else {
+    throw Exception("no select list");
```
lmwnshn, line 44: Generally speaking, I prefer putting the error cases (which tend to be shorter) before the actual logic. This makes it easier to keep track of what's going on IMO.

```
if (blah == nullptr) {
  // error case 1
} else if (blablabla) {
  // error case 2
} else {
  // logic
}
```

IMO, this has the added advantage that if you need to de-indent by one level, it is easy to refactor without breaking anything since the checks are already in the right order.

lmwnshn (changes requested):
This is the first time I'm reviewing a substantial PR from you so I'm more nitpicky than usual. In the future, I'd probably LGTM most of the below without further comment.

Generally LGTM, no logic or architectural changes requested, only minor nitpicks about docstrings, forward declarations, using declarations, and style. Quick notes about the binder tests and modifying core semantics for tests.

**Docstrings**: unfortunately `//` is not picked up by doxygen (which is in theory our standard) and also not picked up by some IDE's hover functionality. Please use `/** */` for class members, `/**< */` for inline enum comments. Have suggested fixes for most of what I saw, probably missed one or two. 

**Forward declarations**: this is probably trauma from trying to cut the compile time in other codebases. As a general guideline, you can include whatever you want in a cpp file, you can include any stdlib stuff (e.g., `<memory>`, `<utility>`) in a header file, but be very careful when including header files in other header files and prefer forward declaration where possible. Otherwise compile time will blow up some day, when someone includes `boost` for a small check and ends up including it in basically every file. Even though BusTub is "just" an educational codebase, it sets our coding style guidelines and expectations for future courses and research codebases, so it would be helpful if we can indoctrinate people into practicing include discipline early. ~~something something modules~~

**Using declarations**: I think it is better to be explicit for students. For example, `using` something like `Exception` makes it really vague as to what kind of `Exception` it is without digging around the codebase, and this might conflict with a future change where someone wants to `using` a different type of `Exception`. We also end up with at least two instances where we have both the short and long forms in the same file, e.g., `std::move()` and `move()` on the same line. I am in favor of removing most `using` declarations that just hide namespacing. 

**Style**:
- Please avoid relative path imports.
- Please avoid or define abbreviations in things like class header files that students are likely to read (e.g., op, agg). The average student may have never written SQL before this course, assume very little background knowledge.
- Please be consistent with naming where possible (ExecutorContext -> GetExecutorContext). 
- Additionally, I find it helpful to assign consistent meaning to words like `Get` and `Make`. `GetBlah` implies you are getting an existing `Blah`, low-cost operation. `MakeBlah` implies that you are creating a new `Blah`, potentially but not necessarily expensive. 
- If you have very short cases in an if-else, please prefer putting the short cases first. Usually, this means that error cases go on top. This helps people to retain context by the time they reach the else block.
- I'm probably the only person who keeps nitpicking about this, so you can probably ignore this if you don't care that much / disagree: please try to write comments like sentences (capitalize the first word, end with fullstop). I find that it makes the codebase look more consistent and "cared for", for lack of a better term.

**Binder tests**: I am not a fan of tests that just print something because nobody checks. I am also not a fan of exact-match tests that just checks for a hardcoded string, but arguably this is more useful / "test-like" because it'll actually break the test suite if the binder breaks. I suggest we move to exact-match tests in lieu of you having to test individual aspects of the bound statements.

**Modifying core semantics for tests**: There is one place where we added `if (== nullptr)` in core code, to support test code. I am generally wary of doing stuff like that, I think it gets out of control very quickly. Generally, I would like to keep core code as assertable as possible (e.g., if you can assert something is not null, you should do it).

Overall, nice PR!

skyzh:
Thanks for the detailed review! This is really helpful :)

src/include/common/config.h
skyzh: Yes, I'll revert this and set buffer pool size for the shell.

src/binder/statement/select_statement.cpp
skyzh: Indeed this is one of the most significant change of this PR 🤣 I'll apply best practice to this file.

skyzh:
Partially resolved the comments. Still WIP.

# Regarding fmt and header usage

For `fmt`: I think it's a utility like `printf` and `iostream`, and putting it in `.h` makes it easier for people to call them. Also implement fmt for base class requires placing the declaration in header. Therefore, I'd prefer keep it in `.h`.

Also fmt provides two headers: `core.h` and `format.h`. I replaced all usage of fmt to `core.h` in headers.

This PR didn't increase compile time too much for clang (only +5s ~ +10s in CI). However, gcc compilation is slower (-30s). As we're already suggesting students using clang, we can ignore the compilation time difference.

For other headers, I've cleaned up a little bit and now we're importing as fewer headers as possible.

# Regarding `std::xxx`

We're using `std::xxx` everywhere now.

skyzh:
GitHub PR review web interface is a disaster if we have too many comments. I'd reply to them as normal PR comments below. ⬇️

skyzh:
For `BUSTUB_ASSERT`: I've fixed in this PR. There're many other occurrences of `assert` in other files, I'll have a separate PR to fix them.

src/include/common/constants.h
```diff
@@ -45,19 +45,8 @@ namespace bustub {
...
-static constexpr int INVALID_TXN_ID = -1;                                     // invalid transaction id
-static constexpr int INVALID_LSN = -1;                                        // invalid log sequence number
-static constexpr int HEADER_PAGE_ID = 0;                                      // the header page id
-static constexpr int PAGE_SIZE = 4096;                                        // size of a data page in byte
-static constexpr int BUFFER_POOL_SIZE = 10;                                   // size of buffer pool
-static constexpr int LOG_BUFFER_SIZE = ((BUFFER_POOL_SIZE + 1) * PAGE_SIZE);  // size of a log buffer in byte
-static constexpr int BUCKET_SIZE = 50;                                        // size of extendible hash bucket
-static constexpr uint64_t COLUMN_IDENTIFIER_ROW_ID = static_cast<uint64_t>(-1);
-
 // size of extendible hash bucket
```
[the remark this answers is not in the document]
skyzh: it turned out that nobody is use this constant, so I'll remove it.

src/include/binder/table_ref/bound_base_table_ref.h
```diff
@@ -0,0 +1,23 @@
...
+
+#include <fmt/format.h>
+#include <string>
+#include <utility>
+#include "../bound_table_ref.h"
+
+namespace bustub {
+
+/**
+ * A bound table ref type for single table. e.g., `SELECT x FROM y`, where `y` is `BoundBaseTableRef`.
```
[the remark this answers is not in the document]
skyzh: We don't have BaseTableRef :)

src/common/bustub_instance.cpp
skyzh: It is a test fixture indeed for executor tests. For this `GenerateTestTable`, it's used in the shell to predefine some tables, as we don't support create / drop table and insert for now.

skyzh:
> Generally speaking, I prefer putting the error cases (which tend to be shorter) before the actual logic. This makes it easier to keep track of what's going on IMO.

skyzh:
> Generally speaking, I prefer putting the error cases (which tend to be shorter) before the actual logic. This makes it easier to keep track of what's going on IMO.

What about `switch`? For most of the times, the binder assumes `default` as an unsupported parse node. Is it okay to make `default` section the first of the switch clause, or just leave it as the last?

src/binder/statement/select_statement.cpp
skyzh: > Generally speaking, I prefer putting the error cases (which tend to be shorter) before the actual logic. This makes it easier to keep track of what's going on IMO.
> 
> ```
> if (blah == nullptr) {
>   // error case 1
> } else if (blablabla) {
>   // error case 2
> } else {
>   // logic
> }
> ```
> 
> IMO, this has the added advantage that if you need to de-indent by one level, it is easy to refactor without breaking anything since the checks are already in the right order.

It turns out that it's due to a clang-tidy lint:

```
/Users/skyzh/Work/bustub/src/binder/statement/select_statement.cpp:48:5: error: do not use 'else' after 'throw' [readability-else-after-return,-warnings-as-errors]
  } else {
    ^~~~~~
```

clang-tidy thinks it's better not to place return / throw branches before actual work. 🤣 I prefer following the clang-tidy rules.

skyzh:
@lmwnshn I've fixed/replied to nearly all of the comments. PTAL, thanks!

src/binder/statement/select_statement.cpp
lmwnshn: That rule is actually to reduce the level of indentation, AFAIK. 

```
if (error checking) {
   ...
}
```

Then you add the code without the else.

lmwnshn:
> I assume it's an intermediate stage when we're to introduce planner and binder. If we can run SQL end-to-end, we can remove binder test.

Ok, please add a TODO comment.

> What about switch? For most of the times, the binder assumes default as an unsupported parse node. Is it okay to make default section the first of the switch clause, or just leave it as the last?

I think it is fine to leave a switch default as the last.

---

Thanks for addressing the comments, will take a second look later.

src/common/bustub_instance.cpp
lmwnshn: OK

src/common/bustub_instance.cpp
lmwnshn: OK, could you please add a TODO?

src/include/binder/bound_table_ref.h
```diff
@@ -0,0 +1,64 @@
...
+
+  virtual auto ToString() const -> std::string {
+    switch (type_) {
+      case TableReferenceType::INVALID:
+        return "<invalid>";
+      case TableReferenceType::EMPTY:
+        return "<empty>";
+      default:
+        // for other types of table reference, `ToString` should be derived in child classes.
+        assert(false && "entered unreachable code");
```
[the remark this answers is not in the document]
lmwnshn: Per your other comment, can fix later.

src/include/binder/table_ref/bound_base_table_ref.h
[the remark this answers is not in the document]
lmwnshn: 5am brain..

src/common/bustub_instance.cpp
lmwnshn: Well, no super strong opinion here.

src/include/common/config.h
```diff
@@ -27,6 +27,7 @@ extern std::atomic<bool> enable_logging;
 /** If ENABLE_LOGGING is true, the log should be flushed to disk every LOG_TIMEOUT. */
 extern std::chrono::duration<int64_t> log_timeout;
 
+static constexpr const uint64_t INVALID_INDEX = -1;
```
lmwnshn, line 30: -1 for a `uint64_t`, arguably using `std::numeric_limits` is a better idea

src/binder/fmt_impl.cpp
```diff
@@ -0,0 +1,6 @@
+#include "binder/expressions/bound_agg_call.h"
+#include "fmt/format.h"
+
+namespace bustub {
+auto BoundAggCall::ToString() const -> std::string { return fmt::format("{}({})", func_name_, fmt::join(args_, ", ")); }
```
lmwnshn, line 5: If this is the only one that you moved to .cpp, maybe we can move this back to .h?

src/include/common/config.h
skyzh: It's also not used anywhere, I'll check for refsol and remove it. Seems this is added when I attempt to merge a file from DuckDB.

lmwnshn (approved):
LGTM, added a few more comments but I'm happy with the overall PR, nice work.

Issue #254, opened 2022-07-04
Question about ReaderWriterLatch


alkaid0815:
Is it correct to add ` if (reader_count_ == MAX_READERS - 1)` ? 
For example, there are two threads called Rlock() blocked in `reader_.wait(latch)` because of `reader_count_ == MAX_READERS`. then the RUnlock() is called by another thread, `if (reader_count_ == MAX_READERS - 1) {
        reader_.notify_one();
      }` this if statement is true, then notify_one(), however, notify_one is just only awake one blocked thread(means it could be scheduled by scheduler, not immediately switch to the waiting thread and continue execute the wait()). So if there is another RUnlock() call before the wait() is called, then the ` if (reader_count_ == MAX_READERS - 1)` is false, so there will be a missing call to `reader_.notify_one()`. The result is only one thread is awaked but it should be two threads being awaked.

```c++
/**
   * Acquire a read latch.
   */
  void RLock() {
    std::unique_lock<mutex_t> latch(mutex_);
    while (writer_entered_ || reader_count_ == MAX_READERS) {
      reader_.wait(latch);
    }
    reader_count_++;
  }

  /**
   * Release a read latch.
   */
  void RUnlock() {
    std::lock_guard<mutex_t> guard(mutex_);
    reader_count_--;
    if (writer_entered_) {
      if (reader_count_ == 0) {
        writer_.notify_one();
      }
    } else {
      if (reader_count_ == MAX_READERS - 1) {
        reader_.notify_one();
      }
    }
  }
```

skyzh:
Looks reasonable to me! I'm thinking of a fix...

skyzh:
But I think it's rare case that we will hit `MAX_READERS` -- it's `UINT_MAX` for now. I believe this bug is easier to be reproduced if we set `MAX_READERS` to a smaller value...

Issue #247, opened 2022-05-31
starter_test.cpp cannot catch ExceptionType in InitializationTest


ctxdata:
./test/starter_test will crash after enabling InitializationTest.

```
unknown file: Failure
Unknown C++ exception thrown in the test body.
[  FAILED  ] StarterTest.InitializationTest (33137 ms)
[ DISABLED ] StarterTest.DISABLED_ElementAccessTest
[ DISABLED ] StarterTest.DISABLED_AdditionTest
[ DISABLED ] StarterTest.DISABLED_MultiplicationTest
[----------] 2 tests from StarterTest (33137 ms total)

[----------] Global test environment tear-down
[==========] 2 tests from 1 test suite ran. (33137 ms total)
[  PASSED  ] 1 test.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] StarterTest.InitializationTest
```

The root cause is ThrowsBustubException won't catch ExceptionType, it catches `const Exception&` instead. While in RowMatrix<T>::FillFrom specification, implementor must throw ExceptionType::OUT_OF_RANGE if the size is mismatched.

To fix the issue, we should catch ExceptionType in ThrowsBustubException

```
catch (const ExceptionType &et) {
    expected_type_thrown = et == type;
  }
```

skyzh:
2021 Fall's p0 is not maintained any more. We'll remove it soon.

PR #300, opened 2022-08-29
chore(docs): remove previous project p0; add comments for rwlatch

skyzh:

close https://github.com/cmu-db/bustub/issues/247
close https://github.com/cmu-db/bustub/issues/254 as won't fix


src/include/common/rwlatch.h
```diff
@@ -80,6 +80,9 @@ class ReaderWriterLatch {
       }
     } else {
       if (reader_count_ == MAX_READERS - 1) {
+        // TODO: this implementation will lead to deadlock after reader count reaches
```
lmwnshn, line 83:
```suggestion
        // TODO(Chi): this implementation will lead to deadlock after reader count reaches
```

We normally include someone in the TODO parens as, "if we need more information, this is the person who'd probably know"

lmwnshn (approved):
LGTM, suggested one change.

PR #306, opened 2022-08-30
chore(build): discover tools after being installed

skyzh:
Students won't need to remove the build directory, though this is recommended.


skyzh:
Short explanation: we should always find_program as `CLANG_FORMAT_BIN` is cached. Once the program is not found, it will be set to `CLANG_TIDY_BIN-NOTFOUND` and be cached. The second invocation won't trigger find_program again.

PR #307, opened 2022-08-31
feat(planner): plan and execute select table and simple agg

skyzh:
This PR adds a simple planner, which plans table scan and agg. Example in shell:

image


skyzh:
Tests will be added after planning where + group by + having is finished. For now let's get the trivial planner framework merged.

skyzh:
Let me add more docs 😋

lmwnshn:
I'll probably review this tomorrow morning.

src/include/planner/planner.h
```diff
@@ -0,0 +1,73 @@
...
+
+  auto PlanExpression(const BoundExpression &expr, const AbstractPlanNode &child)
+      -> std::unique_ptr<AbstractExpression>;
+
+  /** the root plan node of the plan tree */
+  std::unique_ptr<AbstractPlanNode> plan_;
+
+ private:
+  // The collection of allocated things, owned by the planner.
+  // TODO(chi): refactor ALL OF THE ARGS TO UNIQUE_PTR/SHARED_PTR instead of guessing about the ownership.
```
lmwnshn, line 44: I personally really liked the [ManagedPointer](https://github.com/cmu-db/noisepage/blob/master/src/include/common/managed_pointer.h) that we had in NoisePage, which is a little like observer_ptr (which afaik is still experimental).

I think someone suggested that managed pointer might be too much for students, I personally think it might be ok but it is up to you.

That said, I'm not a fan of shared_ptr for most use cases, would prefer a rule of unique ptr to denote ownership + managed or raw pointer to denote non-ownership.

src/include/planner/planner.h
```diff
@@ -0,0 +1,73 @@
...
+
+  auto SavePlanNode(std::unique_ptr<AbstractPlanNode> plan_node) -> const AbstractPlanNode * {
+    allocated_plan_nodes_.emplace_back(move(plan_node));
```
lmwnshn, line 47:
```suggestion
    allocated_plan_nodes_.emplace_back(std::move(plan_node));
```

src/include/planner/planner.h
```diff
@@ -0,0 +1,73 @@
...
+    return allocated_plan_nodes_.back().get();
+  }
+
+  auto SaveSchema(std::unique_ptr<Schema> schema) -> const Schema * {
+    allocated_output_schemas_.emplace_back(move(schema));
```
lmwnshn, line 52:
```suggestion
    allocated_output_schemas_.emplace_back(std::move(schema));
```

src/include/planner/planner.h
```diff
@@ -0,0 +1,73 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "binder/bound_statement.h"
+#include "binder/statement/select_statement.h"
```
lmwnshn, line 10: I feel like we can remove a few of these.. but the planner is hopefully not included in very many places, so no strong opinion.

src/planner/planner.cpp
```diff
@@ -0,0 +1,157 @@
...
+namespace bustub {
+
+void Planner::PlanQuery(const BoundStatement &statement) {
+  switch (statement.type_) {
+    case StatementType::SELECT_STATEMENT: {
+      plan_ = PlanSelect(dynamic_cast<const SelectStatement &>(statement));
+      return;
+    }
+    default:
+      throw Exception("the statement is not supported in planner yet");
```
lmwnshn, line 19: Can you use fmt to print the statement type in the error message? If it is printable.

src/planner/planner.cpp
```diff
@@ -0,0 +1,157 @@
...
+        output_schema.emplace_back(column.GetName(),
+                                   SaveExpression(std::make_unique<ColumnValueExpression>(0, idx, column.GetType())));
+        idx += 1;
+      }
+      return std::make_unique<SeqScanPlanNode>(SaveSchema(MakeOutputSchema(output_schema)), nullptr, table->oid_);
+    }
+    default:
+      break;
+  }
+  throw Exception("the table ref type is not supported in planner yet");
```
lmwnshn, line 154: Same comment as above, print type if possible

lmwnshn (approved):
Minor comments; check if you meant to commit the txn in bustub instance; the rest are mostly a few style things and the overall PR LGTM

src/binder/statement/select_statement.cpp
```diff
@@ -204,6 +204,7 @@ auto SelectStatement::ResolveColumn(const std::string &col_name) -> std::unique_
   switch (table_->type_) {
     case TableReferenceType::BASE_TABLE: {
       auto base_table_ref = dynamic_cast<BoundBaseTableRef *>(table_.get());
+      // TODO(chi): handle case-insensitive table / column names
```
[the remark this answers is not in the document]
skyzh: The reason here is Postgres parser will automatically convert all identifiers to lower-case. But our tables are generated with programs instead of SQLs, so their table names are upper-lower-mixed. We'll need to convert all table names stored internally to lowercase when binding the parse tree.

src/common/bustub_instance.cpp
```diff
@@ -46,14 +56,61 @@ auto BustubInstance::ExecuteSql(const std::string &sql) -> std::vector<std::stri
...
+    planner.PlanQuery(*statement);
+    std::cerr << "=== PLANNER ===" << std::endl;
+    std::cerr << planner.plan_->ToString() << std::endl;
+
+    // Execute the query.
+    auto txn = transaction_manager_->Begin();
+    auto exec_ctx = MakeExecutorContext(txn);
+    std::vector<Tuple> result_set{};
+    execution_engine_->Execute(planner.plan_.get(), &result_set, txn, exec_ctx.get());
+    delete txn;
```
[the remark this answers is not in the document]
skyzh: We don't support any write operations (e.g. insert, update) for now in the shell. Added TODO here. Will add commit / abort when we have support for binding inserts.

src/include/planner/planner.h
skyzh: Yes, that's why I don't care much about forward declaration for this file 🤣 But I'll still try forwarding decl more.

src/binder/statement/select_statement.cpp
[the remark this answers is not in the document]
lmwnshn: ah, sorry, I was going to mark this as resolved / it was a cheap jab at mysql

src/common/bustub_instance.cpp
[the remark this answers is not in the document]
lmwnshn: OK, sounds good

Issue #288, opened 2022-08-27
Proposal: introduce ThreadSanitizer to bustub


skyzh:
Currently we've already enabled AddressSanitizer + LeakSanitizer. We can further add a CI step / grading step to run all tests under thread sanitizer, so as to find potential issues with deadlocks in the bustub project.

skyzh:
One useful thing about thread sanitizer is that it can find potential deadlocks and problems with atomic variables. e.g. if there are two mutex used in two threads, where each thread acquires the mutex in different order, thread sanitizer will report a warning.

GavinRay97:
One question -- I could not find how it is that `LeakSanitizer` is currently enabled, is this part of ASAN by default?

Also, if making changes, another to consider might be to swap `-O0` for `-Og` in debug mode:

> If you are not using some other optimization option, consider using -Og (see [Optimize Options](https://gcc.gnu.org/onlinedocs/gcc/Optimize-Options.html#Optimize-Options)) with -g.

![image](https://user-images.githubusercontent.com/26604994/187085391-df2452d0-885b-4c4a-a328-487d560ab40b.png)

skyzh:
Yes, it's part of ASAN.

PR #310, opened 2022-09-02
chore(build): support thread sanitizer

skyzh:
Thread sanitizer can help check potential race conditions. By default, thread sanitizers print warnings. We can modify it to exit with code 1 if any race condition is detected in the future.

Instructions on enabling thread sanitizer is described in readme.

macOS users will need to use AppleClang to build thread sanitizer. So I'm going to change some descriptions in README and cmake build scripts to recommend macOS users to use AppleClang instead of homebrew clang if they want to use thread sanitizer.

close https://github.com/cmu-db/bustub/issues/288

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -84,10 +84,17 @@ endif ()
 # COMPILER SETUP
 ######################################################################################################################
 
+if(NOT BUSTUB_SANITIZER)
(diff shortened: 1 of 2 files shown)


lmwnshn (approved):
LGTM

PR #311, opened 2022-09-02
chore(common): use shared_mutex for rwlatch

skyzh:
Fewer bugs, more happiness. Merge after p0 so that students won't be affected by this change.


lmwnshn (approved):
Heh. LGTM

PR #314, opened 2022-09-06
refactor(container): separate memory and disk hash table

skyzh:

Still see if CI could pass...


skyzh:
I didn't sync extendible_hash_table_test.cpp (which needs adding `DISABLED_` prefix to all test cases). May do later. Also need to revisit what we want to expose to students for `extendible_hash_table.h`.

skyzh:
For reviewers: the only new files in this PR is src/container/hash/extendible_hash_table.cpp and src/include/container/hash/extendible_hash_table.h. All other files are just about renaming, but git apparently doesn't recognize such renames.

skyzh:
Ready for review, needs approval from @timlee0119

timlee0119 (approved):
LGTM!

PR #324, opened 2022-09-08
feat(binder): bind join and cross join

skyzh:

In this PR, we added support for bind `from A inner join B on Condition` and `from A, B where Condition`.

### DIFF
--- a/src/include/binder/table_ref/bound_join_ref.h
+++ b/src/include/binder/table_ref/bound_join_ref.h
@@ -47,7 +47,7 @@ class BoundJoinRef : public BoundTableRef {
   /** The right side of the join. */
   std::unique_ptr<BoundTableRef> right_;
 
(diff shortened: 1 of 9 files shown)


src/include/binder/table_ref/bound_join_ref.h
```diff
@@ -0,0 +1,82 @@
...
+  /** Type of join. */
+  JoinType join_type_;
+
+  /** The left side of the join. */
+  std::unique_ptr<BoundTableRef> left_;
+
+  /** The right side of the join. */
+  std::unique_ptr<BoundTableRef> right_;
+
+  /** Join condition */
```
lmwnshn, line 50:
```suggestion
  /** Join condition. */
```

lmwnshn (approved):
LGTM

PR #325, opened 2022-09-08
chore(container): throw on unimplemented functions

skyzh:

I spent ~20mins debugging the bustub shell, and realized that the strange ASAN error is caused by unimplemented extendible hash tables. Therefore, we throw exceptions here. Students will also know that they should work on hash table first before working on buffer pools if they see the exceptions.


lmwnshn (approved):
LGTM

PR #326, opened 2022-09-08
chore(type): use snprintf instead of sprint

### LINKED ISSUE #322: timestamp_type.cpp should use `snprintf`, not `sprintf`
`sprintf` is deprecated on recent MacOS SDKs. The `// NOLINT` directives won't prevent a compilation warning (and thus an error). Since the since of the buffers being written to is known, these uses should be changed to `snprintf` instead, which is safer.

peterb12:
timestamp_type.cpp should use snprintf, not sprintf #322. This change allows building on recent versions of MacOS, and should be safer besides.

close #322

### DIFF
--- a/src/type/timestamp_type.cpp
+++ b/src/type/timestamp_type.cpp
@@ -114,11 +114,11 @@ auto TimestampType::ToString(const Value &val) const -> std::string {
   auto day = static_cast<uint16_t>(tm % 32);
   tm /= 32;
   auto month = static_cast<uint16_t>(tm);
-  const size_t dateStrLen = 30;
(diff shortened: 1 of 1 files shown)


skyzh (approved):
Personally I'd prefer using libfmt `fmt::format("{}{}{}{}", arg1, arg2, arg3, arg4)` something like that. But I'm fine with this change. LGTM!

skyzh:
Please fix clang-tidy, thanks!

peterb12:
This is oppression against weFew, weProud, weHappy camelCase bandsOfBrothers!

:-) Fixed.

PR #327, opened 2022-09-09
feat(planner): plan join

skyzh:

PTAL if you're interested and if you have time :) Comments welcomed @TennyZhuang @leiysky

### DIFF
--- a/src/common/bustub_instance.cpp
+++ b/src/common/bustub_instance.cpp
@@ -32,7 +32,7 @@ BustubInstance::BustubInstance(const std::string &db_file_name) {
   // Log related.
   log_manager_ = new LogManager(disk_manager_);
 
-  buffer_pool_manager_ = new BufferPoolManagerInstance(BUFFER_POOL_SIZE, disk_manager_, LRUK_REPLACER_K, log_manager_);
+  buffer_pool_manager_ = new BufferPoolManagerInstance(128, disk_manager_, LRUK_REPLACER_K, log_manager_);
 
   // Transaction (txn) related.
   lock_manager_ = new LockManager();
--- a/src/common/bustub_instance.cpp
+++ b/src/common/bustub_instance.cpp
@@ -32,6 +32,8 @@ BustubInstance::BustubInstance(const std::string &db_file_name) {
   // Log related.
   log_manager_ = new LogManager(disk_manager_);
 
+  // We need more frames for GenerateTestTable to work. Therefore, we use 128 instead of the default
+  // buffer pool size specified in `config.h`.
   buffer_pool_manager_ = new BufferPoolManagerInstance(128, disk_manager_, LRUK_REPLACER_K, log_manager_);
 
   // Transaction (txn) related.
--- a/src/include/catalog/column.h
+++ b/src/include/catalog/column.h
@@ -77,7 +77,7 @@ class Column {
   auto IsInlined() const -> bool { return column_type_ != TypeId::VARCHAR; }
 
   /** @return a string representation of this column */
-  auto ToString() const -> std::string;
+  auto ToString(bool simplified = true) const -> std::string;
 
   /** @return the expression used to create this column */
   auto GetExpr() const -> const AbstractExpression * { return expr_; }
--- a/src/planner/planner.cpp
+++ b/src/planner/planner.cpp
@@ -1,14 +1,23 @@
 #include "planner/planner.h"
 #include "binder/bound_statement.h"
 #include "binder/expressions/bound_agg_call.h"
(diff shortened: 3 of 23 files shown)


skyzh:
Some thoughts:

* Shall we infer schema when constructing plan node, or in planner?
* Does `ColumnValueExpression` look good? Currently it is `child_id.column_id` instead of simply column_id.
* We should make all expressions / schema / plan node into `std::shared_ptr<const AbstractXxxx>`.
* Shall we push-down predicate for cross-join in planner? Or leave a filter plan node here and wait for a future optimizer to do this :(

src/planner/planner.cpp
```diff
@@ -25,18 +34,56 @@ void Planner::PlanQuery(const BoundStatement &statement) {
...
     case ExpressionType::AGG_CALL:
       throw Exception("agg call should be handled by PlanSelect");
     case ExpressionType::COLUMN_REF: {
-      auto column_ref_expr = dynamic_cast<const BoundColumnRef &>(expr);
-      auto schema = child.OutputSchema();
-      // TODO(chi): differentiate multiple tables
-      uint32_t col_idx = schema->GetColIdx(column_ref_expr.col_);
-      auto col_type = schema->GetColumn(col_idx).GetType();
-      return std::make_unique<ColumnValueExpression>(0, col_idx, col_type);
+      const auto &column_ref_expr = dynamic_cast<const BoundColumnRef &>(expr);
```
leiysky, line 43: Why does `COLUMN_REF` expression have children? IMO, it should be the leaf node.

skyzh: If an expression is for join condition, e.g.

SELECT * from test_1 inner join test_2 on test_1.colA = test_2.col2

The plan will be like:

NestedLoopJoin condition = ColumnRef 0.0 = ColumnRef 1.2
  SeqScan colA, colB
  SeqScan col1, col2

where child_id is, when executor is using the expression, it picks from its nth children's mth column to get the data.

src/common/bustub_instance.cpp
```diff
@@ -32,7 +32,7 @@ BustubInstance::BustubInstance(const std::string &db_file_name) {
   // Log related.
   log_manager_ = new LogManager(disk_manager_);
 
-  buffer_pool_manager_ = new BufferPoolManagerInstance(BUFFER_POOL_SIZE, disk_manager_, LRUK_REPLACER_K, log_manager_);
+  buffer_pool_manager_ = new BufferPoolManagerInstance(128, disk_manager_, LRUK_REPLACER_K, log_manager_);
```
lmwnshn, line 35: I remember you mentioned needing to change this; please document hardcoded constants (in a nearby comment above is fine)

src/include/catalog/column.h
```diff
@@ -77,7 +77,7 @@ class Column {
   auto IsInlined() const -> bool { return column_type_ != TypeId::VARCHAR; }
 
   /** @return a string representation of this column */
-  auto ToString() const -> std::string;
+  auto ToString(bool simplified = true) const -> std::string;
```
lmwnshn, line 80: I think Andy generally does not prefer having default arguments and historically they have been a bit of an issue, but I don't feel very strongly about it in this specific use-case. Just a FYI

src/planner/planner.cpp
lmwnshn: I think this explanation is helpful, can you include a comment for context about what the `size == 1` and `size == 2` branches are assuming? (and fix the 1.2 -> 1.1) It can be a very short description, it's just that students might get scared off by trying to read the C++ directly without comments to break it up a bit

lmwnshn (approved):
LGTM, just some minor comments

src/include/catalog/column.h
skyzh: I think we can use the simplified representation as the default one, and move the previous ToString to `ToStringVerbose`.

src/planner/planner.cpp
skyzh: updated :)

src/include/catalog/column.h
skyzh: I'll change this later in batch, revisiting all such occurrences and see if we can simplify `ToString` of other classes.

skyzh:
Fixing clang-tidy 🙀

skyzh:
Bad news: the formatter API for enum formatting we're using is in `format.h` instead of `core.h`. The issue appears today due to some magics in C++. I'll have to use `format.h` everywhere. Hope it doesn't slow down to much on compile time (which doesn't have much impact on compile time on my MacBook)

PR #333, opened 2022-09-11
feat(binder): support bind create table

skyzh:

This PR also refactors the binder, so that the actual binding process will happen in binder instead of the constructor of `SelectStatement`.

Added support for binding bool expr.

Added `NewContext` for managing the current binding scope, and we don't need to pass `scope` in every function.


src/include/binder/binder.h
```diff
@@ -37,36 +37,64 @@
...
 
 #include "binder/simplified_token.h"
+#include "binder/statement/create_statement.h"
 #include "catalog/column.h"
+#include "common/macros.h"
 #include "nodes/parsenodes.hpp"
 #include "type/type_id.h"
 #include "type/value.h"
 
+#include "binder/binder.h"
```
lmwnshn, line 47: Maybe I'm not seeing this correctly, we are including ourself?

Most of these includes could probably be fwd decls, granted I think there shouldn't be too many things (transitively) importing binder.h

lmwnshn (approved):
LGTM with one qn about recursive include?

src/include/binder/binder.h
skyzh: You're write, I'm surprised it's not detected!

I'll do forward decl.

PR #334, opened 2022-09-11
feat(optimizer): add noop optimizer

skyzh:

After a second thought, I think it would be a better idea to do something in a separate phase instead of in planner. Examples:

* Select index for join
* Select join algorithm (rewrite NLJ to HashJoin)
* Filter push-down
* ...

### DIFF
--- a/src/common/bustub_instance.cpp
+++ b/src/common/bustub_instance.cpp
@@ -94,7 +94,7 @@ auto BustubInstance::ExecuteSql(const std::string &sql) -> std::vector<std::stri
 
     // Optimize the query.
     bustub::Optimizer optimizer(*catalog_);
-    auto optimized_plan = optimizer.Optimize(&*planner.plan_);
+    auto optimized_plan = optimizer.Optimize(planner.plan_.get());
     std::cerr << "=== OPTIMIZER ===" << std::endl;
     std::cerr << optimized_plan->ToString() << std::endl;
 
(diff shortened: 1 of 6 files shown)


src/common/bustub_instance.cpp
```diff
@@ -91,11 +92,17 @@ auto BustubInstance::ExecuteSql(const std::string &sql) -> std::vector<std::stri
     std::cerr << "=== PLANNER ===" << std::endl;
     std::cerr << planner.plan_->ToString() << std::endl;
 
+    // Optimize the query.
+    bustub::Optimizer optimizer(*catalog_);
+    auto optimized_plan = optimizer.Optimize(&*planner.plan_);
```
lmwnshn, line 97:
```suggestion
    auto optimized_plan = optimizer.Optimize(planner.plan_.get());
```

If this works, I think it would be clearer than `&*`

src/include/optimizer/optimizer.h
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <memory>
+#include <string>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "catalog/catalog.h"
+#include "execution/plans/abstract_plan.h"
```
lmwnshn, line 10: Forward declare?

lmwnshn (approved):
Generally LGTM

PR #336, opened 2022-09-11
feat(binder): bind and plan alias

skyzh:

```
SELECT colA as a from __mock_table_1
```

now can be correctly planned. The output column will also have `a` name.

The issue is, we cannot do planning like:

```
SELECT colA as a from __mock_table_1 where a > 10
```

That's because `where` planning is using the scope of `from` clause (__mock_table_1), which absolutely doesn't have the a column. We should make `SELECT colA as a from __mock_table_1` a kind of table function when binding and planning.

If we make `select` a table function, we can also easily do uncorrelated subquery. I'm thinking whether we can do this in fall 2022.

### DIFF
--- a/src/binder/bind_select.cpp
+++ b/src/binder/bind_select.cpp
@@ -4,6 +4,7 @@
 #include "binder/bound_expression.h"
 #include "binder/bound_statement.h"
 #include "binder/expressions/bound_agg_call.h"
+#include "binder/expressions/bound_alias.h"
 #include "binder/expressions/bound_binary_op.h"
 #include "binder/expressions/bound_column_ref.h"
 #include "binder/expressions/bound_constant.h"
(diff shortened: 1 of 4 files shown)


skyzh:
... maybe not the case

```
select sum(v1) as sum from t1 where sum > 10;
```

doesn't work in DuckDB, but

```
select v1 as v3, sum(v2) from t2 where v3 > 10 group by v3;
```

works... So we cannot treat `select <list> from table` as a subquery before bind where.

skyzh:
> Standard SQL disallows references to column aliases in a WHERE clause. This restriction is imposed because when the WHERE clause is evaluated, the column value may not yet have been determined.

It's not allowed in standard! We don't need to do this :)

lmwnshn:
It certainly is very convenient though.

iwanttobepowerful:
image
image

lmwnshn (approved):
LGTM

PR #338, opened 2022-09-11
feat(binder): bind insert and values

skyzh:

* Refactor planner and split `plan_table_ref.cpp` out.
* Plan `values` as `ExpressListTableRef`
* Bind insert.

### DIFF
--- a/src/binder/statement/insert_statement.cpp
+++ b/src/binder/statement/insert_statement.cpp
@@ -1,10 +1,29 @@
+#include "fmt/ranges.h"
+
+#include "binder/bound_expression.h"
+#include "binder/bound_table_ref.h"
 #include "binder/statement/insert_statement.h"
+#include "binder/statement/select_statement.h"
+#include "common/util/string_util.h"
 
 namespace bustub {
 
-InsertStatement::InsertStatement(duckdb_libpgquery::PGInsertStmt *pg_stmt)
-    : BoundStatement(StatementType::INSERT_STATEMENT) {}
+auto IndentAllLines(const std::string &lines, size_t num_indent) -> std::string {
+  std::vector<std::string> lines_str;
+  auto lines_split = StringUtil::Split(lines, '\n');
+  lines_str.reserve(lines_split.size());
+  auto indent_str = StringUtil::Indent(num_indent);
+  for (auto &line : lines_split) {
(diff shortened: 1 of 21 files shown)


src/include/binder/bound_expression.h
```diff
@@ -31,7 +31,7 @@ class BoundExpression {
   BoundExpression() = default;
   virtual ~BoundExpression() = default;
 
-  virtual auto ToString() const -> std::string { return "<invalid>"; };
+  virtual auto ToString() const -> std::string { return ""; };
```
lmwnshn, line 34: Any particular reason for this change?

src/binder/statement/insert_statement.cpp
```diff
@@ -1,10 +1,29 @@
...
+#include "binder/bound_table_ref.h"
 #include "binder/statement/insert_statement.h"
+#include "binder/statement/select_statement.h"
+#include "common/util/string_util.h"
 
 namespace bustub {
 
-InsertStatement::InsertStatement(duckdb_libpgquery::PGInsertStmt *pg_stmt)
-    : BoundStatement(StatementType::INSERT_STATEMENT) {}
+auto IndentAllLines(const std::string &lines, size_t num_indent) -> std::string {
```
lmwnshn, line 11: This seems like code that belongs in StringUtil

lmwnshn (approved):
LGTM, one suggested refactor + personally I preferred when invalid would be explicitly rendered as such, but if there's a reason for the change we can leave it

src/include/binder/bound_expression.h
skyzh: So that students can be less confusing about the binder output. Previously, if we don't have where clause, it will output `<invalid>` as if the clause is not correctly parsed. We should print nothing instead.

skyzh: I'm thinking of remove `invalid` and use `std::optional` / `nullptr` to represent a non-existing branch. Thinking about it...

src/binder/statement/insert_statement.cpp
skyzh: +1

skyzh:
Or maybe we should remove `invalid` and make it something like `empty`. It would be better, probably. Invalid statements should already be rejected in binder.

PR #340, opened 2022-09-11
feat(planner): plan insert

skyzh:
As title

### DIFF
--- a/src/planner/plan_insert.cpp
+++ b/src/planner/plan_insert.cpp
@@ -1,0 +1,19 @@
+#include <memory>
+
+#include "binder/statement/insert_statement.h"
+#include "execution/expressions/column_value_expression.h"
+#include "execution/plans/insert_plan.h"
+#include "planner/planner.h"
+
+namespace bustub {
+
+auto Planner::PlanInsert(const InsertStatement &statement) -> std::unique_ptr<AbstractPlanNode> {
+  auto select = PlanSelect(*statement.select_);
+
+  auto table = catalog_.GetTable(statement.table_);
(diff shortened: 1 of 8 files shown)


src/planner/plan_insert.cpp
```diff
@@ -0,0 +1,19 @@
...
+
+namespace bustub {
+
+auto Planner::PlanInsert(const InsertStatement &statement) -> std::unique_ptr<AbstractPlanNode> {
+  auto select = PlanSelect(*statement.select_);
+
+  auto table = catalog_.GetTable(statement.table_);
+  BUSTUB_ASSERT(table, "table not found");
+
+  return std::make_unique<InsertPlanNode>(SavePlanNode(std::move(select)), table->oid_);
```
lmwnshn, line 16: Is the other Insert ctor (taking in a `vec<vec<values>> raw_values`) handled somewhere? This assumes that the insert statement is an INSERT INTO ... SELECT?

lmwnshn (approved):
Clarifying question, LGTM

src/planner/plan_insert.cpp
skyzh: It won't be used anymore after we integrated the new frontend. I'll remove it once we've merged all PRs.

skyzh: Even for `INSERT INTO ... VALUES`, values will be planned as `ValuesExecutor`. Within `ValuesExecutor`, it stores `Vec<Vec<AbstractExpression>>`. We will never directly generate `Value` in the query processing layer.

PR #346, opened 2022-09-11
feat(planner): plan hash agg

skyzh:
As title, we can plan `group by` now.

Having is not support for now. Select group by column is not supported for now (we can only select aggregation functions).

### DIFF
--- a/src/execution/fmt_impl.cpp
+++ b/src/execution/fmt_impl.cpp
@@ -37,7 +37,8 @@ static auto HelperVecExprFmt(const std::vector<const AbstractExpression *> &expr
 
 auto AggregationPlanNode::PlanNodeToString() const -> std::string {
   auto aggregates = HelperVecExprFmt(aggregates_);
-  return fmt::format("Agg {{ types={}, aggregates=[{}] }}", agg_types_, aggregates);
+  return fmt::format("Agg {{ types={}, aggregates=[{}], having={}, group_by=[{}] }}", agg_types_, aggregates,
+                     having_ == nullptr ? "" : having_->ToString(), HelperVecExprFmt(group_bys_));
 }
 
 auto ProjectionPlanNode::PlanNodeToString() const -> std::string {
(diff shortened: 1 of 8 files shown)


skyzh:
A more complex hashagg like:

```
select v3, max(v1) + max(v2) from table group by v3 having count(v4) > count(v5);
```

will be handled later.

src/execution/fmt_impl.cpp
```diff
@@ -37,7 +37,8 @@ static auto HelperVecExprFmt(const std::vector<const AbstractExpression *> &expr
 
 auto AggregationPlanNode::PlanNodeToString() const -> std::string {
   auto aggregates = HelperVecExprFmt(aggregates_);
-  return fmt::format("Agg {{ types={}, aggregates=[{}] }}", agg_types_, aggregates);
+  return fmt::format("Agg {{ types={}, aggregates=[{}], having={}, group_by=[{}] }}", agg_types_, aggregates,
+                     having_ == nullptr ? "" : having_->ToString(), HelperVecExprFmt(group_bys_));
```
lmwnshn, line 41: IMO explicitly representing nullptr as "nullptr" is clearer, but up to you

lmwnshn: Same comment for the other places this happens.

Actually, the `blah == nullptr ? "" : blah->ToString()` is a common enough pattern that maybe it can be made a macro or util function somewhere. But that might be overdoing things

lmwnshn (approved):
LGTM

src/execution/fmt_impl.cpp
skyzh: Exactly! After I've made everything into `shared_ptr / unique_ptr`, I'll change this to:

```
fmt::format("having={}", having_)
```

We will have implementation of format `unique_ptr<Expression>`, which yields "nullptr" or "" automatically, without manually handling the case.

PR #347, opened 2022-09-12
refactor: sync from private repo

skyzh:
cc @timlee0119 we missed one function from the private repo. Shall we add it to the public header? (They can remove it at any time).

I was scared when I see the header of `src/storage/index/b_plus_tree.cpp`, but it turned out it isn't a private one. So I removed the header.

### DIFF
--- a/src/include/container/hash/extendible_hash_table.h
+++ b/src/include/container/hash/extendible_hash_table.h
@@ -172,6 +172,12 @@ class ExtendibleHashTable : public HashTable<K, V> {
   mutable std::mutex latch_;
   std::vector<std::shared_ptr<Bucket>> dir_;  // The directory of the hash table
 
+  /**
+   * @brief Redistribute the kv pairs in a full bucket.
+   * @param bucket The bucket to be redistributed.
+   */
+  auto RedistributeBucket(std::shared_ptr<Bucket> bucket) -> void;
+
   /*****************************************************************
    * Must acquire latch_ first before calling the below functions. *
    *****************************************************************/
--- a/src/include/container/hash/extendible_hash_table.h
+++ b/src/include/container/hash/extendible_hash_table.h
@@ -165,12 +165,16 @@ class ExtendibleHashTable : public HashTable<K, V> {
   };
 
  private:
-  // TODO(student): You may add additional private members and helper functions
+  // TODO(student): You may add additional private members and helper functions and remove the ones
+  // you don't need.
+
   int global_depth_;    // The global depth of the directory
   size_t bucket_size_;  // The size of a bucket
   int num_buckets_;     // The number of buckets in the hash table
   mutable std::mutex latch_;
   std::vector<std::shared_ptr<Bucket>> dir_;  // The directory of the hash table
+
+  // The following functions are completely optional, you can delete them if you have your own ideas.
 
   /**
    * @brief Redistribute the kv pairs in a full bucket.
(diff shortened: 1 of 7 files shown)


src/include/container/hash/extendible_hash_table.h
```diff
@@ -172,6 +172,12 @@ class ExtendibleHashTable : public HashTable<K, V> {
   mutable std::mutex latch_;
   std::vector<std::shared_ptr<Bucket>> dir_;  // The directory of the hash table
 
+  /**
+   * @brief Redistribute the kv pairs in a full bucket.
+   * @param bucket The bucket to be redistributed.
+   */
+  auto RedistributeBucket(std::shared_ptr<Bucket> bucket) -> void;
```
timlee0119, line 179: This is a private helper method for the ref solution so I don't think we should push it to public. Is it not working well with hacky_sync?

skyzh: I put it here with intention. 🤣 As we didn't give the implementation of this function, I think it would be fine.

The bad side of ignoring this file from synchronizing between two repos are that, if you make any change in the public repo, you will need to carefully merge it with private repo, and it will go out-of-sync very soon. There're a lot of examples in this PR, like we have `const` in private repo but not in public, etc. So I think it would be better to have a smaller set of ignored files.

timlee0119: We probably want to put this file in the .syncignore file to not sync it? And only expose the public apis of extendible hash table in the public repo

timlee0119: Make sense! Let's sync it then haha

lmwnshn: > Also this hints the students "you might want to implement this function to make your life easier". If they don't want it, they can remove it.

Can you please add this as a comment above the function, for any such function? Even something like `// Optional, feel free to delete.`

skyzh: It's already in the beginning of the private section:

```
 private:
  // TODO(student): You may add additional private members and helper functions
```

I'll make it more detailed 🤤

test/CMakeLists.txt
```diff
@@ -33,6 +33,8 @@ foreach (bustub_test_source ${BUSTUB_TEST_SOURCES})
             --gtest_color=yes
             --gtest_output=xml:${CMAKE_BINARY_DIR}/test/${bustub_test_name}.xml
             --gtest_catch_exceptions=0
+            PROPERTIES
+            TIMEOUT 30
```
lmwnshn, line 37: Is this what is now `DISCOVERY_TIMEOUT` or is this test timeout? If the former, OK (but arguably we should use the new form instead, as of `3.10.3` it got renamed to avoid this confusion). If the latter, how/where will we handle tests > 30s?

lmwnshn (approved):
Generally LGTM, please address comments

test/CMakeLists.txt
skyzh: This is test timeout. Test cases > 30s will be considered failure if invoked using `ctest`. If invoked directly by the binary (e.g., ./test/binder_test), then it doesn't take effect.

src/include/container/hash/extendible_hash_table.h
lmwnshn: Change LGTM, thanks! Now it should be very clear

PR #350, opened 2022-09-15
feat(planner): plan having and select group by column in hashagg

skyzh:
Not tested with refsol yet, but the plan looks correct. See comments in `PlanSelectAgg` for more.

```
bustub> select colB, sum(colA) > max(colB) from __mock_table_1 group by colB having min(colA) == 3;
=== BINDER ===
BoundSelect {
  table=BoundBaseTableRef { table=__mock_table_1 },
  columns=[__mock_table_1.colB, (sum([__mock_table_1.colA])>max([__mock_table_1.colB]))],
  groupBy=[__mock_table_1.colB],
  having=(min([__mock_table_1.colA])==3),
  where=
}
=== PLANNER ===
Projection { exprs=[#0.0, #0.2>#0.3] } | (__mock_table_1.colB:INTEGER, <unnamed>:BOOLEAN)
  Filter { predicate=#0.1=3 } | (__mock_table_1.colB:INTEGER, agg#0:INTEGER, agg#1:INTEGER, agg#2:INTEGER)
    Agg { types=[min, sum, max], aggregates=[#0.0, #0.0, #0.1], having=, group_by=[#0.1] } | (__mock_table_1.colB:INTEGER, agg#0:INTEGER, agg#1:INTEGER, agg#2:INTEGER)
      MockScan { size=100 } | (__mock_table_1.colA:INTEGER, __mock_table_1.colB:INTEGER)
=== OPTIMIZER ===
Projection { exprs=[#0.0, #0.2>#0.3] } | (__mock_table_1.colB:INTEGER, <unnamed>:BOOLEAN)
  Filter { predicate=#0.1=3 } | (__mock_table_1.colB:INTEGER, agg#0:INTEGER, agg#1:INTEGER, agg#2:INTEGER)
    Agg { types=[min, sum, max], aggregates=[#0.0, #0.0, #0.1], having=, group_by=[#0.1] } | (__mock_table_1.colB:INTEGER, agg#0:INTEGER, agg#1:INTEGER, agg#2:INTEGER)
      MockScan { size=100 } | (__mock_table_1.colA:INTEGER, __mock_table_1.colB:INTEGER)
```

### DIFF
--- a/src/planner/plan_select.cpp
+++ b/src/planner/plan_select.cpp
@@ -1,0 +1,67 @@
+#include <memory>
+#include <utility>
+
+#include "binder/bound_expression.h"
+#include "binder/bound_statement.h"
+#include "binder/bound_table_ref.h"
+#include "binder/statement/insert_statement.h"
+#include "binder/statement/select_statement.h"
+#include "binder/tokens.h"
+#include "common/enums/statement_type.h"
+#include "common/exception.h"
+#include "common/macros.h"
+#include "common/util/string_util.h"
+#include "execution/expressions/abstract_expression.h"
+#include "execution/expressions/column_value_expression.h"
+#include "execution/plans/abstract_plan.h"
+#include "execution/plans/filter_plan.h"
+#include "execution/plans/projection_plan.h"
+#include "fmt/format.h"
+#include "planner/planner.h"
+
+namespace bustub {
+
+auto Planner::PlanSelect(const SelectStatement &statement) -> std::unique_ptr<AbstractPlanNode> {
+  std::unique_ptr<AbstractPlanNode> plan = nullptr;
+
+  switch (statement.table_->type_) {
+    case TableReferenceType::EMPTY:
+      throw Exception("select value is not supported in planner yet");
+    default:
+      plan = PlanTableRef(*statement.table_);
+      break;
+  }
+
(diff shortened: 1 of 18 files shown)


skyzh:
cc @st1page do you feel like this is a better way to do such planning (compared with RisingLight / RisingWave?)

src/planner/plan_select.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+  switch (statement.table_->type_) {
+    case TableReferenceType::EMPTY:
+      throw Exception("select value is not supported in planner yet");
+    default:
+      plan = PlanTableRef(*statement.table_);
+      break;
+  }
+
+  if (!statement.where_->IsInvalid()) {
```
yliang412, line 35: Curious question, how does it handle the case that the `where` clause is invalid?

yliang412 (approved):
LGTM, great work!

src/planner/plan_select.cpp
skyzh: When users don't specify `where` clause, it will be invalid. So we don't handle it. e.g.,

```
select * from table;
```

In this case, `where` will be invalid in BoundSelect.

I agree that `invalid` might be confusing here. I'm thinking of changing it to `IsEmpty()`.

PR #366, opened 2022-09-17
fix: release mode build

skyzh:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,6 +10,11 @@ project(BusTub
         LANGUAGES C CXX
         )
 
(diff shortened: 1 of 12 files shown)


skyzh:
I'll also add MemoryDiskManager in this PR, just trying to squash more things...

timlee0119 (approved):
looks great:)

skyzh:
We'll run cmake in debug mode if no build type is specified from now on (e.g., `cmake ..`).

Also we disabled warning for release mode, so that for `BUSTUB_ASSERT` variables, clang won't complain about unused variables.

Issue #344, opened 2022-09-11
Proposal: remove raw pointers in plan nodes


skyzh:
We use `AbstractPlanNode *` and `AbstractExpression *` in these places. This makes it harder to infer lifetime of these variables. We use hacks like `SaveExpression` and `SavePlanNode` in planner to ensure they have at least the same lifetime of the planner.

I propose to change all of them to `std::shared_ptr<const AbstractPlanNode>` and `std::shared_ptr<const AbstractExpression>`, so as to make our life easier.

PR #373, opened 2022-09-19
feat(style): use smart pointer in planner

skyzh:
close https://github.com/cmu-db/bustub/issues/344


skyzh:
This PR consists of several changes.

* Use shared_ptr for AbstractExpression, AbstractPlanNode, and Schema when possible.
* Remove `AbstractExpression` in `Column`. Now we have projection plan node, and we don't need every executor to do some kind of projection again.
* Refactor fmt implementation so that we can directly provide `vector<shared_ptr<T>>` to the formatter. No need to manually format or check nullptr.
* Remove `EvaluateAggregate` on expressions. No need any more.

Some remaining works:

* In some places, we still mix the use of raw pointers and references. In some places, we convert references to pointers. We will refactor them later.
* Insert / Delete / Update executor should have a schema.

skyzh:
This PR introduces a bug when planning aggregation reporting "unexpected agg" in some circumstances. The fix is in https://github.com/cmu-db/bustub/pull/374 and I don't want to do a merge 🤣

yliang412 (approved):
LGTM

PR #380, opened 2022-09-20
fix(planner): check insert schema match

skyzh:
close https://github.com/cmu-db/bustub/issues/375


skyzh:
insert / delete / update executor should return number of lines updated in the future.

yliang412 (approved):
LGTM

Issue #362, opened 2022-09-16
planner: plan create index


skyzh:

skyzh:
postponed to the next year as personally we don't want students to implement too many executors for p3

PR #387, opened 2022-09-24
feat(binder): bind index and delete

skyzh:
This PR adds support for binding `create index` and `delete from table <where x>`.

close https://github.com/cmu-db/bustub/issues/359
close https://github.com/cmu-db/bustub/issues/362


yliang412 (approved):
LGTM

PR #388, opened 2022-09-25
feat(optimizer): plan order by as index scan, support create index

skyzh:

* Some refactors on binder, now we bind full table info for insert
* Support create index statement
* Support plan delete
* Fix some bugs in optimizer
* Support plan order by index column as index scan


yliang412 (approved):
LGTM

PR #389, opened 2022-09-25
B+ Tree: Strip out implementation-specific functions

durovo:
1. Strip out implementation-specific functions
2. Add function for retrieving root page id of tree
3. Add test for some remaining public APIs


test/storage/b_plus_tree_insert_test.cpp
```diff
@@ -39,6 +39,49 @@ TEST(BPlusTreeTests, DISABLED_InsertTest1) {
   auto header_page = bpm->NewPage(&page_id);
   (void)header_page;
 
+  int64_t key = 42;
+  int64_t value = key & 0xFFFFFFFF;
+  rid.Set(static_cast<int32_t>(key >> 32), value);
```
skyzh, line 44: `key >> 32` is zero? If so why do we set key = 42 and do some "transformations" instead of directly `int key = 0`?

test/storage/b_plus_tree_insert_test.cpp
```diff
@@ -39,6 +39,49 @@ TEST(BPlusTreeTests, DISABLED_InsertTest1) {
...
+  tree.Insert(index_key, rid, transaction);
+
+  auto root_page_id = tree.GetRootPageId();
+  auto root_page = reinterpret_cast<BPlusTreePage *>(bpm->FetchPage(root_page_id)->GetData());
+  ASSERT_NE(root_page, nullptr);
+  ASSERT_TRUE(root_page->IsLeafPage());
+
+  auto root_as_leaf = reinterpret_cast<BPlusTreeLeafPage<GenericKey<8>, RID, GenericComparator<8>> *>(root_page);
+  ASSERT_EQ(root_as_leaf->GetSize(), 1);
+  ASSERT_EQ(comparator(root_as_leaf->KeyAt(0), index_key), 0);
```
skyzh, line 55: hmm, so it's zero 🤤

test/storage/b_plus_tree_insert_test.cpp
```diff
@@ -39,6 +39,49 @@ TEST(BPlusTreeTests, DISABLED_InsertTest1) {
   auto header_page = bpm->NewPage(&page_id);
   (void)header_page;
```
skyzh, line 40: Can we assert header_page_id == 0 here?

src/include/storage/page/b_plus_tree_internal_page.h
```diff
@@ -40,27 +40,9 @@ class BPlusTreeInternalPage : public BPlusTreePage {
 
   auto KeyAt(int index) const -> KeyType;
   void SetKeyAt(int index, const KeyType &key);
```
skyzh, line 42: It seems that the API is not paired. e.g., we have APIs to read key and set key, but we only have API to read value without setting value... Can you think of any better way of stripping this part?

skyzh (approved):
Rest LGTM, good work!

test/storage/b_plus_tree_insert_test.cpp
durovo: Good catch. This is an artifact from the previous test1 from which this was copied.

test/storage/b_plus_tree_insert_test.cpp
durovo: The actual value used will be the RID which is constructed using CONCAT(key>>32, value) I believe, so it should not be zero. In this case, this is using the comparator to compare the key, not the value. This should be 42 I believe.

src/include/storage/page/b_plus_tree_internal_page.h
durovo: Hmm, this is one area where I feel adding a Set operation might be confusing. Especially in the leaf page where I feel that a SetKey operation does not make sense, but GetKey does.

skyzh:
Looks good to me. May merge if you think it's ready.

PR #391, opened 2022-09-26
feat(expr): arithmetic and logic expression

skyzh:

* Support `select 1, 2, 3`.
* Also fixes a bug on subquery, so that we can `select * from (select 1, 2, 3)`.

close https://github.com/cmu-db/bustub/issues/356
close https://github.com/cmu-db/bustub/issues/357


yliang412 (approved):
Great work

PR #401, opened 2022-09-29
Update package.sh. On MacOS, If you already have other versions of git, no longer install via brew.

grejioh:
If you already have other versions of git, no longer install via brew.
Most mac os systems have git pre-installed (apple git), I don't think installing git via brew is necessary. Should we use git --version to verify that git is already installed?

### DIFF
--- a/build_support/packages.sh
+++ b/build_support/packages.sh
@@ -79,7 +79,7 @@ install_mac() {
   brew ls --versions cmake || brew install cmake
   brew ls --versions coreutils || brew install coreutils
   brew ls --versions doxygen || brew install doxygen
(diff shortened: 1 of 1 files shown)


skyzh:
Power users like you should customize your installation and figure how whether you need to install these tools for themselves. I think it's okay to install more than install less (which leads to students complaining on Piazza).

skyzh:
I would probably close this PR. Feel free to reopen 🤣

grejioh:
> I would probably close this PR. Feel free to reopen 🤣

I understand, thanks for the reply. ❤️

Issue #382, opened 2022-09-21
demo: BusTub reference implementation in browser


skyzh:
After some easy fixes of compiler errors in https://github.com/cmu-db/bustub/pull/381, we can now run BusTub in WASM. But there're more things to do before we can use it in browser.

* Redirect textbox to stdin. WASM doesn't seem to have native support for std::cin.
* Web UI for SQL.

After that, we can let students try our reference implementation in browser.

Issue #371, opened 2022-09-17
Proposal: pre-compiled bustub-shell for students


skyzh:
In the future, we can release pre-compiled bustub-shell with our reference solutions, so that students can get a vague idea about how the complete version of BusTub looks like before they finish all projects.

PR #407, opened 2022-09-30
feat(build): add web shell

skyzh:
This PR adds WASM shell, and fix several bugs / add several improvements in the system.

* Explain support options (show only binder / planner / optimizer result)
* values executor should set cursor to 0 when init
* merge projection rule is wrong

close https://github.com/cmu-db/bustub/issues/382
close https://github.com/cmu-db/bustub/issues/371


yliang412 (approved):
Bustub in browser!!!

Issue #398, opened 2022-09-27
expr: bind `and` `or` correctly


skyzh:
Currently, if we use something like `a > 10 and a > 20 and a > 30`, it seems that the parser will parse it as `and` with 3 children, which is not supported by BusTub now. We should fix this in the planner.

Issue #397, opened 2022-09-27
bug: binder cannot handle multiple dependent DDL in one input


skyzh:
We parse and bind all statements within one SQL input, and then plan. This causes problems.

PR #408, opened 2022-10-01
fix(binder): allow multiple DDL in one parsed statement, fix and/or

skyzh:
close https://github.com/cmu-db/bustub/issues/398
close https://github.com/cmu-db/bustub/issues/397


yliang412 (approved):
LGTM

PR #425, opened 2022-10-07
fix(tools): flush when draw to file, stop when cin is invalid

skyzh:

### DIFF
--- a/tools/b_plus_tree_printer/b_plus_tree_printer.cpp
+++ b/tools/b_plus_tree_printer/b_plus_tree_printer.cpp
@@ -25,10 +25,10 @@ using bustub::DiskManager;
 using bustub::Exception;
 using bustub::GenericComparator;
 using bustub::GenericKey;
-using bustub::RID;
-using bustub::Transaction;
 using bustub::page_id_t;
 using bustub::ParseCreateStatement;
+using bustub::RID;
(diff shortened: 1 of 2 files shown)


tools/b_plus_tree_printer/b_plus_tree_printer.cpp
```diff
@@ -84,6 +84,9 @@ auto main(int argc, char **argv) -> int {
   while (!quit) {
     std::cout << "> ";
     std::cin >> instruction;
+    if (!std::cin) {
```
durovo, line 87: It seems like invalid inputs are being explicitly handled in the switch block. Was this change introduced to handle some different failure scenario?

durovo (approved):

tools/b_plus_tree_printer/b_plus_tree_printer.cpp
skyzh: If you press Ctrl+D in the current shell, it will infinitely loop. We use `std::cin as bool` to check if we reached the end of the stdin. This is not the same as handling invalid inputs.

skyzh: ... and you're correct! This is used to handle some different failure scenario!

PR #435, opened 2022-10-18
feat(executor): support left joins, fix tuple bug

yliang412:
This PR supports left-join for nested loop join, nested index join, and hash join. It also fixes some memory issues for null values in the `Tuple` type.


skyzh (approved):
LGTM!

Issue #450, opened 2022-10-29
[build_support/packages.sh] Ruby Homebrew installer disabled on macOS 12 Monterey


ancientmodern:
## Environment
A new MacBook Pro (m1pro) with macOS 12.5 Monterey and homebrew is not installed before.

## Log
```shell
ancientmodern@MacBook-Pro$ build_support/packages.sh
PACKAGES WILL BE INSTALLED. THIS MAY BREAK YOUR EXISTING TOOLCHAIN.
YOU ACCEPT ALL RESPONSIBILITY BY PROCEEDING.
Proceed? [Y/n] : Y
++ uname
++ tr '[:lower:]' '[:upper:]'
+ UNAME=DARWIN
+ case $UNAME in
+ install_mac
++ which brew
+ test '!'
+ echo 'Installing Homebrew (https://brew.sh/)'
Installing Homebrew (https://brew.sh/)
++ curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install
+ ruby -e '#!/usr/bin/ruby

abort <<EOS
Error: The Ruby Homebrew installer is now disabled and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"

EOS'
Error: The Ruby Homebrew installer is now disabled and has been rewritten in
Bash. Please migrate to the following command:
  /bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"
```

Then I manually ran `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"` and `build_support/packages.sh`, and everything went well. Seems that the `ruby -e "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/master/install)"`should be replaced by `/bin/bash -c "$(curl -fsSL https://raw.githubusercontent.com/Homebrew/install/HEAD/install.sh)"`

ancientmodern:
I open PR #451 for this, but I do not know if personal PR is accepted :)

PR #451, opened 2022-10-29
fix(build): migrate Ruby Homebrew Install to Bash

ancientmodern:
Ruby Homebrew Install is now disabled and has been rewritten in Bash #450. This change allows macOS users without installing Homebrew before to set up the dev environment correctly. `eval ...` is added in order to add `brew` to PATH immediately, otherwise the remaining part will fail.


skyzh (dismissed):

skyzh:
Would you please if it works on x86? Only M1 Mac installs homebrew at `/opt`.

PR #452, opened 2022-10-30
[DNM] feat(p2): add index based on std::map

skyzh:

as a workaround for those who cannot finish project 2

### DIFF
--- a/src/include/storage/index/b_plus_tree.h
+++ b/src/include/storage/index/b_plus_tree.h
@@ -76,12 +76,18 @@ class BPlusTree {
   void Print(BufferPoolManager *bpm) { throw bustub::NotImplementedException("not implemented"); }
 
   // draw the B+ tree
-  void Draw(BufferPoolManager *bpm, const std::string &outf);
+  void Draw(BufferPoolManager *bpm, const std::string &outf) {
+    throw bustub::NotImplementedException("not implemented");
+  }
 
   /* Debug Routines for FREE!! */
-  void ToGraph(BPlusTreePage *page, BufferPoolManager *bpm, std::ofstream &out) const;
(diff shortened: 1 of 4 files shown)


skyzh:
This PR should not be merged. Anyone who needs it can pull from this branch.

src/include/storage/index/stl_comparator_wrapper.h
```diff
@@ -0,0 +1,14 @@
+#pragma once
+
+namespace bustub {
+
+template <typename KeyType, typename KeyComparator>
+class StlComparatorWrapper {
+ public:
+  explicit StlComparatorWrapper(const KeyComparator &cmp) : cmp_{cmp} {}
+
+  inline auto operator()(const KeyType &lhs, const KeyType &rhs) const -> bool { return cmp_(lhs, rhs) < 0; }
```
lmwnshn, line 10: Oh! I just realized why my wrapper on Friday was seeing weird bugs...

I thought `operator()` would return `int` like a Java comparator, not a `bool` based on `std::less`. So I just used `cmp_(lhs, rhs)` directly. Oops.

skyzh: Yeah, I also feel C++'s compile error is unhelpful and cannot point to the exact place that I got wrong with 😇

sacloudy:
Does the autograder server contain this header file? When I pull this branch and submit it to autograder, I got this error: 
/autograder/source/bustub/src/include/storage/index/index_iterator.h:20:10: fatal error: 'storage/index/stl_comparator_wrapper.h' file not found

skyzh:
> Does the autograder server contain this header file? When I pull this branch and submit it to autograder, I got this error:
> /autograder/source/bustub/src/include/storage/index/index_iterator.h:20:10: fatal error: 'storage/index/stl_comparator_wrapper.h' file not found

See the Discord discussion.

PR #460, opened 2022-11-02
chore(p3): make fields of plan node public

skyzh:

So as to make life easier for optimizer hackers :)


yliang412 (approved):
LGTM

PR #462, opened 2022-11-02
fix: allow old copy of optimizer

skyzh:

Remove this after p3 deadline and before p4 🤣


yliang412 (approved):
LGTM

PR #472, opened 2022-11-16
chore(p3): use in-memory backend

skyzh:
Leaderboard tests are not affected. Only test cases invoked with ctest will use in-memory backend.


yliang412 (approved):
LGTM

PR #487, opened 2022-11-26
fix(interface): modified ExtendibleHashTable::Insert() return type

AkiyamaKunka:
From CMU 15-445 2022 fall [Project Webpage](https://15445.courses.cs.cmu.edu/fall2022/project1/#buffer-pool-instance), `ExtendibleHash::Insert()` method need return a `bool` type instead of `void`.

image

### DIFF
--- a/src/container/hash/extendible_hash_table.cpp
+++ b/src/container/hash/extendible_hash_table.cpp
@@ -75,7 +75,7 @@ auto ExtendibleHashTable<K, V>::Remove(const K &key) -> bool {
 }
 
 template <typename K, typename V>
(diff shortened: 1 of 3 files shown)


odysa:
Could you please also update the comment?
https://github.com/cmu-db/bustub/blob/25f8d91fbc602f956928be36ed26ef420b2bd42a/src/include/container/hash/extendible_hash_table.h#L84-L95

skyzh:
I guess this will affect existing projects for this semester. Postpone merging to next semester. LGTM, thanks!

skyzh:
fixed with https://github.com/cmu-db/bustub/pull/524

PR #489, opened 2022-12-07
fix(p3): correctly the tuple index of the right key in hash join

wuxiao889:
right side of join tuple index should be 1


skyzh:
For HashJoin the child id doesn't matter. So we set it to 0 anyways.

PR #490, opened 2022-12-20
refactor: remove buffer pool manager instance

skyzh:
BufferPoolManagerInstance is originally used by project 3 when checking if NLJ is implemented correctly by repeatively scanning all the pages. Given now we have a more powerful execution framework we can actually check if NLJ's child is initialized for multiple times, or check how many times the disk pages are read.


test/catalog/catalog_test.cpp
```diff
@@ -14,7 +14,7 @@
 #include <unordered_set>
 #include <vector>
 
-#include "buffer/buffer_pool_manager_instance.h"
```
yliang412, line 17: I think some of the refactor is not synced?

yliang412: Some of the `BufferPoolManagerInstance` stuff is still present.

skyzh: Fixing now...

yliang412 (approved):
LGTM.

LzwY25:
When I submit project1 to gradescope. it requires buffer_pool_manager_instance.h and buffer_pool_manager_instance.cpp, but I can't find these files.
How can I submit project1 after removing buffer pool manager instance?

skyzh:
Try using an earlier commit. Generally we won't be maintaining projects from previous semesters when a new semester begins.

PR #495, opened 2023-01-13
feat(p0): cow trie and upper/lower expr

skyzh:
PR for p0 starter code. All test cases are public for this project.


src/include/primer/trie.h
```diff
@@ -0,0 +1,124 @@
...
+  // Create a TrieNode with no children.
+  TrieNode() = default;
+
+  // Create a TrieNode with some children.
+  explicit TrieNode(std::map<char, std::shared_ptr<const TrieNode>> children) : children_(std::move(children)) {}
+
+  virtual ~TrieNode() = default;
+
+  // Clone returns a copy of this TrieNode. If the TrieNode has a value, the value is copied. The return
+  // type of this function is a shared_ptr to a TrieNode.
```
yarkhinephyo, line 55:
```suggestion
  // type of this function is a unique_ptr to a TrieNode.
```
Did you mean to write unique_ptr?

src/include/primer/p0_trie.h
```diff
@@ -1,310 +0,0 @@
...
-   * @brief Get the corresponding value of type T given its key.
-   * If key is empty, set success to false.
-   * If key does not exist in trie, set success to false.
-   * If the given type T is not the same as the value type stored in TrieNodeWithValue
-   * (ie. GetValue<int> is called but terminal node holds std::string),
-   * set success to false.
-   *
-   * To check whether the two types are the same, dynamic_cast
-   * the terminal TrieNode to TrieNodeWithValue<T>. If the casted result
-   * is not nullptr, then type T is the correct type.
```
yarkhinephyo, line 298: I realized that this note has been removed here and from the writeup. As a student, tit was really helpful for me. But I understand if the purpose is to challenge them to figure out other (?) ways

skyzh: I think it’s okay to add this back. It is removed because I started from scratch instead of basing on the code from previous semester.

src/include/primer/trie.h
skyzh: Yes, good spot!

src/include/primer/p0_trie.h
skyzh: added it back :)

src/planner/plan_func_call.cpp
```diff
@@ -0,0 +1,45 @@
...
+  }
+  return GetFuncCallFromFactory(expr.func_name_, std::move(args));
+}
+
+// NOLINTNEXTLINE
+auto Planner::GetFuncCallFromFactory(const std::string &func_name, std::vector<AbstractExpressionRef> args)
+    -> AbstractExpressionRef {
+  // 1. check if the function name is lower or upper.
+  // 2. verify the number of args (should be 1).
+  // 3. create `StringExpression`.
```
yarkhinephyo, line 41:
```suggestion
  // 1. check if the parsed function name is "lower" or "upper".
  // 2. verify the number of args (should be 1), refer to the test cases for error msgs.
  // 3. return a `StringExpression` pointer.
```
Just minor nit!

yarkhinephyo (approved):

yliang412 (approved):

src/planner/plan_func_call.cpp
skyzh: I don't think the test case has any requirement for error msgs. I'll change it to `refer to the test cases for when you should throw an error`.

PR #499, opened 2023-01-25
enable color in clang tidy

arvinwu168:
Enables colored error message for clang tidy (used in `make check-clang-tidy-p0`).

- Uses code modified from Django GitHub [repo](https://github.com/django/django/blob/main/django/core/management/color.py) to check if output supports coloring.


skyzh (approved):
LGTM. Will merge after CI passes and I'll get it to the private repo.

PR #502, opened 2023-01-31
refactor(p2): use page guard in B+ tree

abigalekim:
WIP


src/storage/index/b_plus_tree.cpp
```diff
@@ -7,15 +7,39 @@
 #include "storage/page/header_page.h"
 
 namespace bustub {
+
+auto Rp(BufferPoolManager *bpm, page_id_t pid) -> ReadPageGuard {
```
skyzh, line 11: We can call buffer pool manager's FetchPageRead / FetchPageWrite / FetchPageBasic instead.

abigalekim: Done.

build_support/run_clang_tidy.py
```diff
@@ -106,7 +105,7 @@ def get_tidy_invocation(f, clang_tidy_binary, checks, tmpdir, build_path,
     start = [clang_tidy_binary]
 
     if supports_color():
-        start.append('--use-color')
+        start.append('--color')
```
skyzh, line 108: I think this should still be `--use-color` if it's clang-tidy-14 🤪

```
  --use-color                    -
                                   Use colors in diagnostics. If not set, colors
                                   will be used if the terminal connected to
                                   standard output supports colors.
                                   This option overrides the 'UseColor' option in
                                   .clang-tidy file, if any.
```

src/include/storage/page/page_guard.h
```diff
@@ -1,3 +1,5 @@
+
```
skyzh: remove blank lines?

src/storage/index/b_plus_tree.cpp
```diff
@@ -31,7 +32,7 @@ auto BPLUSTREE_TYPE::IsEmpty() const -> bool { return true; }
  * @return : true means key exists
  */
 INDEX_TEMPLATE_ARGUMENTS
-auto BPLUSTREE_TYPE::GetValue(const KeyType &key, std::vector<ValueType> *result, Transaction *transaction) -> bool {
+auto BPLUSTREE_TYPE::GetValue(const KeyType &key, std::vector<ValueType> *result, Transaction *txn) -> bool {
```
skyzh, line 35: I think we should provide more instructions here by defining the prototype of GetHelper / PutHelper in refsol. We should at least show how to use `Context`.

test/CMakeLists.txt
```diff
@@ -28,15 +28,21 @@ foreach (bustub_test_source ${BUSTUB_TEST_SOURCES})
     add_dependencies(build-tests ${bustub_test_name})
     add_dependencies(check-tests ${bustub_test_name})
 
+    SET(TEST_TIMEOUT 120)
+
+    if(${bustub_test_name} STREQUAL "grading_b_plus_tree_checkpoint_2_concurrent_test")
```
skyzh: I would prefer setting a global timeout of i.e. 300s or removing the timeout. The grading script already have timeout for each test case 🤣

test/storage/b_plus_tree_concurrent_test.cpp
```diff
@@ -333,4 +347,73 @@ TEST(BPlusTreeConcurrentTest, DISABLED_MixTest) {
   remove("test.log");
 }
 
+TEST(BPlusTreeConcurrentTest, DISABLED_MixTest2) {
+  // create KeyComparator and index schema
+  auto key_schema = ParseCreateStatement("a bigint");
+  GenericComparator<8> comparator(key_schema.get());
+
+  auto *disk_manager = new DiskManager("test.db");
```
skyzh, line 355: It would be a good idea to replace all DiskManager with InMemoryDiskManagerUnlimited (see project 1 benchmark for more information)

skyzh (approved):
Rest LGTM, good work! Feel free to merge once all comments are resolved.

test/storage/b_plus_tree_concurrent_test.cpp
abigalekim: do you mean DiskManagerUnlimitedMemory()?

src/storage/index/b_plus_tree.cpp
abigalekim: see slack thread, my comment to this contains spoilers about refsol lmao

test/storage/b_plus_tree_concurrent_test.cpp
skyzh: Yes 🤣

src/storage/index/b_plus_tree.cpp
skyzh: Yeah we should put something in the writeup. The refsol didn't implement it in a recursive way 😂. Mark as resolved.

PR #506, opened 2023-02-09
chore: use ptr for page

skyzh:
so that ASAN can detect page overflow (which is rather hard to debug in B+ tree project -- students will find their page pin count magically turned into 0)


abigalekim (approved):
LGTM

PR #515, opened 2023-02-22
fix: do more appropriate space check when inserting a tuple

guomo233:
When inserting a tuple, it will try to find a free slot to reuse, so the space checks at the beginning of the method should not consider the space of the slot. The check of the slot space will be done after the reuse fails.


skyzh:
This part will be refactored this semester, and I guess we won't have such logic any more (as it's prune to delete / update concurrent test failure). I would prefer not merging this for now.

skyzh:
... sorry for that!

skyzh:
fixed with refactor -> https://github.com/cmu-db/bustub/pull/534

PR #523, opened 2023-03-06
chore(p2): move helper functions to new file

skyzh:
Helper functions are causing troubles for MOSS, so it might be better to move them into another file.


skyzh:
we can also postpone this to next semester as it is only causing trouble for only a few of the submissions...

PR #524, opened 2023-03-13
feat: support multi-key indices

skyzh:


skyzh:
Now we have index size of two integers and we can accept maximum of 2-int index columns.

yliang412 (approved):
LGTM

PR #529, opened 2023-03-15
test(p3): add nlj and hash join checks

christopherlim98:
As title says


yliang412 (approved):
Minor suggestions. Remember to update the oh time too

PR #531, opened 2023-03-16
fix: change the dereference of IndexIterator from return-by-reference…

guomo233:
In the multi-threaded case, if a reference is returned without a lock, it may not be safe to read.


skyzh:
We only test iterators in single-thread. Thanks for the suggestion but we are not going to accept this PR for now.

PR #540, opened 2023-03-19
fix: change sql test sources

yliang412:


skyzh (approved):
Would you please reformat using CLion to keep the original style of the file? Otherwise LGTM, good work!

PR #541, opened 2023-03-19
fix(p3): add in place update to avoid dead loop

skyzh:


yliang412 (approved):
Do you also want to just put the Update CMakeLists changes in here?

Issue #309, opened 2022-09-02
Proposal: print backtrace on assert and exception


skyzh:
... so that students can easily figure out how things get wrong instead of getting simply a error message.

infdahai:
perhaps gtests  already provided some suitable backtrace in the repository?

PR #548, opened 2023-03-24
feat: add stack backtrace

infdahai:
issue : #309

just move the [stack trace impl](https://github.com/chenshuo/muduo/blob/master/muduo/base/CurrentThread.cc#L22-L74) in muduo to bustub.


skyzh:
Wow, this is really useful! Thanks a lot!

infdahai:
> Wow, this is really useful! Thanks a lot!

I want to know whether gtests already provided some suitable backtrace in the repository?

infdahai:
> > Wow, this is really useful! Thanks a lot!
> 
> I want to know whether gtests already provided some suitable backtrace in the repository?

I test the gtest output in error situations afterwards.

infdahai:
for example,

**current assert behavior:**
```
[ RUN      ] LockManagerDeadlockDetectionTest.CycleTest1
~/workspace/bustub-private/test/concurrency/deadlock_detection_test.cpp:218: Failure
Expected equality of these values:
  1
  txn_id
    Which is: 4
[  FAILED  ] LockManagerDeadlockDetectionTest.CycleTest1 (50 ms)
```

**expected assert behavior:**
```
[ RUN      ] LockManagerDeadlockDetectionTest.CycleTest1
~/workspace/bustub-private/test/concurrency/deadlock_detection_test.cpp:218: Failure
Expected equality of these values:
  1
  txn_id
    Which is: 4

backtrace:
/workspace/.../a.cpp:100
/workspace/.../b.cpp:50

[  FAILED  ] LockManagerDeadlockDetectionTest.CycleTest1 (50 ms)
```

infdahai:
in mutil-threads test, perhaps we don't get pretty backtrace in `assert failed` situation. it will print the same concurrent func multiple times and get the useless 'lib.so' call info.

src/include/execution/expressions/comparison_expression.h
```diff
@@ -75,6 +76,7 @@ class ComparisonExpression : public AbstractExpression {
         return lhs.CompareGreaterThanEquals(rhs);
       default:
         BUSTUB_ASSERT(false, "Unsupported comparison type.");
+        return CmpBool::CmpFalse;
```
infdahai, line 79: I think it's better not to use assert() as the return value.

skyzh:
after trying this on private repo, it seems that we cannot de-mangle symbols on macOS. Probably I'll prefer wait for another compiler version bump and use https://en.cppreference.com/w/cpp/utility/basic_stacktrace instead in C++ 23.

infdahai:
ok, maybe this situation is related to https://github.com/NationalSecurityAgency/ghidra/issues/4652. so shall i close the pr?

skyzh:
Let’s keep it and I’ll probably see if there’s any chance to integrate it.

Issue #281, opened 2022-08-25
docs: fix license headers and file identifications


skyzh:
The style for header of each file is not consistent across all files in the project. The copyright date isn't updated to 2022. We should figure out a template for all of our file headers and fix this.

This is the style mostly found in this project:

```
//===----------------------------------------------------------------------===//
//
//                         BusTub
//
// rid.h
//
// Identification: src/include/common/rid.h
//
// Copyright (c) 2015-2019, Carnegie Mellon University Database Group
//
//===----------------------------------------------------------------------===//
```

lmwnshn:
I think there is a script somewhere that did this formatting.

skyzh:
Unluckily, the only `.sh` file in this repo is `build_support/packages.sh`. Do you mean some tools outside of this repo?

lmwnshn:
Sorry, I'm bad at checking github notifs. 

I don't remember if we have a newer version than [Peloton](https://github.com/cmu-db/peloton/blob/master/script/formatting/formatter.py), which is basically just a dumb header regex anyway.

I'm inclined to not care about this and punt to the future now that the semester is underway.

PR #549, opened 2023-03-24
[feat] add format_header

infdahai:
this pr target is to solve issue #281 .

1. still incomplete file.


infdahai:
and we should use this header template.

```
//===----------------------------------------------------------------------===//
//
//                         BusTub
//
// rid.h
//
// Identification: src/include/common/rid.h
//
// Copyright (c) 2015-2019, Carnegie Mellon University Database Group
//
//===----------------------------------------------------------------------===//
```

skyzh:
You don't need to replace all headers now. We will need to manually apply it to the private repo. As long as the script works, then I'm happy to get it merged.

infdahai:
oh, I got it.

infdahai:
I guess the format_header py can insert some place then execute the commands like 'make project2'

Issue #404, opened 2022-09-30
RFC: support `select * from 'x.csv'` or `copy from 'x.csv'`


skyzh:
... so that we can test large-scale data with BusTub, like TPC-H.

PR #554, opened 2023-03-28
feat(copy): support `copy from 'x.csv'` grammar

infdahai:
issue: #404

1. just add parser's support for this statement.


infdahai:
notImpl: the `select * from 'a.csv' ` statement is considering where sconst is stored.

skyzh:
I didn't realize `copy from` / `copy to` are not standard Postgres SQL. But given we are actually using DuckDB's parser, I'd prefer keep this repo as-is DuckDB. Do you know which commit you are using from DuckDB?

infdahai:
I will test it in the latest branch after a few days. I forgot the origin commit.

Issue #562, opened 2023-04-16
bug: misues of INTERNAL_PAGE_SIZE in BPlusTreeInternalPage


Roscky:
The definition of INTERNAL_PAGE_SIZE in BPlusTreeInternalPage is correct, but passing INTERNAL_PAGE_SIZE directly in b_plus_tree.h is incorrect because ValueType in BPlusTree is not the same as the ValueType corresponding to BPlusTreeInternalPage. Therefore, the macro definition of INTERNAL_PAGE_SIZE here leads to incorrect calculation results, which can cause wasted space in internal nodes.

The macro definition of INTERNAL_PAGE_SIZE in [b_plus_internal_tree.h](https://github.com/cmu-db/bustub/blob/master/src/include/storage/page/b_plus_tree_internal_page.h).
```
#define INTERNAL_PAGE_SIZE ((BUSTUB_PAGE_SIZE - INTERNAL_PAGE_HEADER_SIZE) / (sizeof(MappingType)))
```
INTERNAL_PAGE_SIZE is used in [b_plus_tree.h](https://github.com/cmu-db/bustub/blob/master/src/include/storage/index/b_plus_tree.h)
```
explicit BPlusTree(std::string name, page_id_t header_page_id, BufferPoolManager *buffer_pool_manager,
                     const KeyComparator &comparator, int leaf_max_size = LEAF_PAGE_SIZE,
                     int internal_max_size = INTERNAL_PAGE_SIZE);
```
Since the ValueType of MappingType in b_plus_tree.h is RID, which is incorrect for InternalPage, the INTERNAL_PAGE_SIZE will be miscalculated.

infdahai:
I think it doesn't cause the situation.  

The macro INTERNAL_PAGE_SIZE is defined in [b_plus_internal_tree.h](https://sourcegraph.com/github.com/cmu-db/bustub@dfa6cd4e82ef42eb111b889604cbf280771b7850/-/blob/src/include/storage/page/b_plus_tree_internal_page.h).

```cpp
// valuetype for internalNode should be page id_t
template class BPlusTreeInternalPage<GenericKey<4>, page_id_t, GenericComparator<4>>;
template class BPlusTreeInternalPage<GenericKey<8>, page_id_t, GenericComparator<8>>;
```
[...]

PR #564, opened 2023-04-19
fix: b+tree internal_page_size

Roscky:
this pr's target is to solve issue #562


skyzh:
Thanks for the PR but I'd prefer a full refactor next semester that removes all such macros...

PR #572, opened 2023-06-07
Update txn_integration_test.cpp

ppdogg:
Add delete test since tools for test already provided.
I think it is ok to just provide delete test A since there are three delete tests online.


skyzh (approved):
LGTM but our intention is that students should recreate these unit tests by themselves... 🤪

PR #573, opened 2023-06-12
fix: modify UnpinPage's annotation in buffer_pool_manager.h

ZXreaper:
When I was doing Project #1 - Buffer Pool, according to the error message of Autograder Results, I found that there may be a problem with the annotation of the UnpinPage method in bustub/src/include/buffer/buffer_pool_manager.h. The annotation mentioned "If page_id is not in the buffer pool, return false", but I can't pass the test of the UnpinPage method in gradescope according to this logic, but if the code logic is changed to "If page_id is not in the buffer pool, return true", it will succeed. I think someone should have encountered this problem.

### DIFF
--- a/src/include/buffer/buffer_pool_manager.h
+++ b/src/include/buffer/buffer_pool_manager.h
@@ -124,8 +124,7 @@ class BufferPoolManager {
   /**
    * TODO(P1): Add implementation
    *
-   * @brief Unpin the target page from the buffer pool. If page_id is not in the buffer pool or its pin count is already
-   * 0, return false.
+   * @brief Unpin the target page from the buffer pool. If its pin count is already 0, return false.
    *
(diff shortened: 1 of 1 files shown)


ZXreaper:
> I don't think pin count will be < 0 in any case. You should revisit your implementation.

Hello, Assistant Professor Chi, I agree with your statement, and the pin count < 0 did not occur when I was doing this project. But I think you seem to have misunderstood what I mean, pin_count<=0 is mentioned in the original annotation, I just deleted the first half sentence in the original annotation, you can check files changed. If I follow the hint in the annotation and return false when page not in page table, I will fail the test on gradescope. 
I have passed all tests according to my logic now. You can check to see if there is such a problem.

ZXreaper:
> Emmm, it seems the result is related to your implementation. The origin doc is right for me.

ok, I will check it, thanks.

prashanthduvvada:
Closing this PR!

PR #582, opened 2023-06-22
feat: Add a private ctor in TableHeap to avoid nullptr table when running binder tests

xzhseh:
To mention, the private ctor will initialize the underlying `bpm_` to `nullptr`, and left the `page_id` to be `INVALID_PAGE_ID`

### DIFF
--- a/src/include/storage/table/table_heap.h
+++ b/src/include/storage/table/table_heap.h
@@ -92,7 +92,17 @@ class TableHeap {
    */
   void UpdateTupleInPlaceUnsafe(const TupleMeta &meta, const Tuple &tuple, RID rid);
 
+  /** For binder tests */
+  static auto CreateEmptyHeap(bool create_table_heap = false) -> std::unique_ptr<TableHeap> {
+    // The input parameter should be false in order to generate a empty heap
+    assert(!create_table_heap);
+    return std::unique_ptr<TableHeap>(new TableHeap(create_table_heap));
+  }
+
  private:
+  /** Used for binder tests */
+  explicit TableHeap(bool create_table_heap = false);
+
   BufferPoolManager *bpm_;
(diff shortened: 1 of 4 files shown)


src/include/storage/table/table_heap.h
```diff
@@ -92,7 +93,17 @@ class TableHeap {
    */
   void UpdateTupleInPlaceUnsafe(const TupleMeta &meta, const Tuple &tuple, RID rid);
 
+  /** For binder tests */
+  static auto CreateEmptyHeap(bool create_table_heap = false) -> std::unique_ptr<TableHeap> {
+    // The input parameter should be false in order to generate a empty heap
+    assert(!create_table_heap);
+    return std::unique_ptr<TableHeap>(new TableHeap(create_table_heap));
```
infdahai, line 100: maybe we prefer to use make_unique.

infdahai:
Perhaps we can just use `CreateEmptyHeap()`. I think the function name is direct.

src/include/storage/table/table_heap.h
xzhseh: Actually I've tried `std::make_unique` before, but seems it can't be apply to private constructor, due to instantiations of the inner templates.

infdahai: Yeah, my mistake. https://stackoverflow.com/questions/65286678/make-unique-cannot-access-private-constructor-in-static-member

yddd123 (approved):

src/include/storage/table/table_heap.h
skyzh: Okay that's interesting... 🤪

skyzh (approved):
Thanks!

PR #592, opened 2023-08-11
fix(test): add check in COW test where root value is written then removed (#590)

skyzh:
cherry-pick as the original PR picks the wrong branch to merge


yliang412 (approved):
LGTM

PR #593, opened 2023-08-14
p0: add new tests + upgrade dependencies

skyzh:
also upgrade to C++ 23 so that we can use std::backtrace


skyzh:
need to revert to C++23 change 🤣

yliang412 (commented):
Does changing to C++23 break other part of the codebase?

xzhseh:
Should we also manually add `-std=c++23` flag in `CMakeLists.txt`🤔

skyzh:
decided to drop C++23 for this semester and bumped several dependencies, PTAL again, thanks! @yliang412

yliang412 (approved):
Great work!

PR #595, opened 2023-08-24
add backward-cpp for backtrace

skyzh:
close https://github.com/cmu-db/bustub/pull/548

This PR adds backtrace for calls to `std::terminate` for test cases. This will be called in a variety of cases, like `BUSTUB_ENSURE` and many other places in the std library. Hopefully this will make debugging easier for students.

On macOS, it cannot show the exact line number of stacktrace due to lack of dwarf support (need libunwind?). On Linux, it will show the full context.


yliang412 (approved):
Great feature to have for the students!

PR #597, opened 2023-08-24
add gradescope agreement and update readme

skyzh:
decide to drop support for Docker / Vagrant


README.md
```diff
@@ -102,66 +103,11 @@ This enables [AddressSanitizer](https://github.com/google/sanitizers) by default
...
-to open a shell within the box. You can find Bustub's code mounted at `/bustub` and run the commands mentioned above like normal.
-
-## Testing
-
-```
-$ cd build
-$ make check-tests
-```
+There are some differences between macOS and Linux (i.e., mutex behavior) that might cause test cases
+to produce difference results in different platforms. We recommend students to use a Linux VM for running
```
xzhseh, line 112:
```suggestion
to produce different results in different platforms. We recommend students to use a Linux VM for running
```

yliang412 (approved):
We probably should annouce Linux/macOS only support together with p0 release.

PR #599, opened 2023-08-29
Feat: improve the interactivity of the b_plus_tree_printer

I-can-do-fourier:
This PR(for macos)aims at improving the interactivity of the b_plus_tree_printer running on our local machine. At present, every time we want to show our result of a bpt, we need to use the command `g my-tree.dot`, then type `dot -Tpng -O my-tree.dot` in another terminal. Finally, we should open/refresh the PNG image to show the results. It wastes a lot of time if we want to examine our implementation step by step. 

This PR adds a new command 's' in the b_plus_tree_printer.cpp. After typing 's', the tree will be stored and converted to the png file. Then the tree image will be open/refresh on the screen. We can split the image to the right side of the screen and print step by step.
p2
p3

### DIFF
--- a/tools/b_plus_tree_printer/b_plus_tree_printer.cpp
+++ b/tools/b_plus_tree_printer/b_plus_tree_printer.cpp
@@ -123,6 +123,15 @@ auto main(int argc, char **argv) -> int {
         std::cin >> filename;
         tree.Draw(bpm, filename);
         break;
+
+      case 's':
+        //std::cin >> filename;
+        tree.Draw(bpm, filename);
+        filename="my-tree.dot";
+        system((std::string("dot -Tpng -O ")+filename).c_str());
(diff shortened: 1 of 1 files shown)


skyzh:
Thanks for the PR but I feel it would be better if students can compile their implementation to WASM so that they can visualize in the browser (as in wasm_bpt_printer)?

I-can-do-fourier:
> Thanks for the PR but I feel it would be better if students can compile their implementation to WASM so that they can visualize in the browser (as in wasm_bpt_printer)?

Thank you, skyzh. I didn't notice there is already a wasm_bpt_printer in our source code. Could you please give me some instructions on how to use this? I try to use `make wasm-bpt-printer -j$(nproc) ` but it fails. I am not very familiar with c++ and cmake.

prashanthduvvada:
I'm closing this issue as the 'wasm_bpt_printer' exists.

PR #610, opened 2023-09-11
p1: use owned promise for disk scheduler

skyzh:
after a second thought it would be a better idea to pass the promise with ownership


fernandolis10 (approved):
LGTM

PR #614, opened 2023-09-21
feat: Add support to `module 'sre_compile' is deprecated`

zizhengWong:
when I run `make check-lint`, I got
```
build_support/cpplint.py:52: DeprecationWarning: module 'sre_compile' is deprecated
import sre_compile
```
this python module has been deprecated

## fix bug
sre_* modules like `sre_constants`, `sre_compile`, and `sre_parse` were deprecated in [3.11](https://github.com/python/cpython/commit/1be3260a90f16aae334d993aecf7b70426f98013)

I use `re._compiler` instead, and it works!

### reference
https://github.com/python/cpython/issues/105456

### DIFF
--- a/build_support/cpplint.py
+++ b/build_support/cpplint.py
@@ -49,7 +49,6 @@ import itertools
 import math  # for log
 import os
 import re
-import sre_compile
 import string
 import sys
 import sysconfig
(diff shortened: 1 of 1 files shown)


prashanthduvvada:
No longer required, hence closing this PR for now.

PR #622, opened 2023-10-12
Feat: add window function support

AveryQi115:
# WIP, DO NOT MERGE

# Major Changes
- [x] binder support for all window functions (partitionBy, orderBy, N preceding/following/unbounded and M preceding/following/unbounded, Rows between/ranges between)
- [x] planner support for window function with default settings (partitionBy, orderBy, Rows between Unbounded and current rows)
- [x] executor starter code
- [x] tests
- [x] clean build errors and add more comments
- [ ] add rank and row numbers


src/include/execution/plans/window_plan.h
```diff
@@ -0,0 +1,135 @@
...
+  BUSTUB_PLAN_NODE_CLONE_WITH_CHILDREN(WindowAggregationPlanNode);
+
+  /** The Window Aggregation column index */
+  int window_agg_id_;
+
+  /** all other columns expressions */
+  std::vector<AbstractExpressionRef> columns_;
+
+  /** The PARTITION BY expressions */
+  std::vector<AbstractExpressionRef> partition_by_;
```
skyzh, line 95: shouldn't this be a vector of "a vector of expressions"? if there are multiple aggregation functions with different partitions and order-by.

src/planner/plan_window_aggregation.cpp
```diff
@@ -0,0 +1,118 @@
...
+  std::vector<std::string> column_names;
+  ;
+  uint32_t window_agg_id = statement.select_list_.size();
+  const BoundWindow *window_call = nullptr;
+
+  for (uint32_t i = 0; i < statement.select_list_.size(); i++) {
+    const auto &item = statement.select_list_[i];
+    if (item->HasWindowFunction()) {
+      if (window_agg_id != statement.select_list_.size()) {
+        throw Exception("multiple window function is not supported yet");
```
AveryQi115, line 47: Currently it only supports one window-function in select list. But I can extend it to support multiple ones.

src/include/execution/plans/window_plan.h
AveryQi115: Currently it only supports one window-function in select list. But I can extend it to support multiple ones, see below.

And I'm thinking about the design of executors, in normal aggregations it is fixed as an initial value first and a combine function to let students implementing the calculation of aggs based on a map. For window agg, are we gonna give similar design or we make it more flexible to define starter code that emits tuples and let students decide their implementation?

skyzh: We can simply given them three empty functions: constructor, init, and next. I’m also planning to strip all code from the current aggregation executor.

src/planner/plan_window_aggregation.cpp
AveryQi115: Multiple window functions supported now

skyzh (approved):
LGTM! would be good if we can add support for commonly-used window functions like ROW_NUMBER and RANK.

AveryQi115:
> LGTM! would be good if we can add support for commonly-used window functions like ROW_NUMBER and RANK.

I'm thinking about add some test queries and test functions first to make sure it won't impact other queries and also make sure window aggregations works fine. Then extend Rank and Row_numbers. And I'll finish this by Wednesday so maybe we can release P3 at wednesday evening?

skyzh:
sounds good :)

test/sql/p3.20-window_function.slt
```diff
@@ -0,0 +1,88 @@
...
+
+# Simple window aggregation over the table
+query
+insert into t1 values (-99999), (99999), (0), (1), (2), (3);
+----
+6
+
+query
+select count(*) over (), min(v1) over (), max(v1) over (), count(v1) over (), sum(v1) over () from t1;
+----
```
skyzh, line 16: do we require order on the output? if not, we can use `query nosort` to skip checking the data order.

PR #627, opened 2023-10-17
chore: add read guard test

skyzh:
an unwritten contract (actually we wrote it in the BPM comment) of page guard


yliang412 (approved):
LGTM.

PR #634, opened 2023-10-25
window aggregation -> window function; change formatter

AveryQi115:


src/binder/bind_select.cpp
```diff
@@ -564,7 +564,7 @@ auto Binder::BindFuncCall(duckdb_libpgquery::PGFuncCall *root) -> std::unique_pt
   }
 
   if (function_name == "min" || function_name == "max" || function_name == "first" || function_name == "last" ||
-      function_name == "sum" || function_name == "count") {
+      function_name == "sum" || function_name == "count" || function_name == "rank") {
```
skyzh, line 567: I think we still need to have ROW_NUMBER here, but rewriting it to count(1) somewhere in planner?

skyzh (approved):
LGTM except that I believe the executor should still be named as `WindowAggregationExecutor`?

src/binder/bind_select.cpp
AveryQi115: I changed it to `WindowFunctionExecutor` because I think rank and row_number are not aggregations and in postgres docs, it is also named as window function.

About Row_number, I'll add the row_number function and change it to count(1) in planner.

PR #658, opened 2023-11-11
p4: more comments and more tests

skyzh:
This pull request adds nc-shell, which allows students to connect to BusTub using netcat. It also completes serializable and abort tests.


skyzh:
image

yliang412 (approved):
LGTM

PR #674, opened 2023-12-24
add request signature script

skyzh:
Students may use `python3 gradescope_sign.py` to generate a signature and we will verify it on Gradescope.


yliang412 (approved):
LGTM!

PR #677, opened 2024-01-05
Update CMakeLists.txt for better debug info

JinjieZhao:
Add -fstandalone-debug to emit full debug info

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,7 @@ message("${BUSTUB_SANITIZER} sanitizer will be enabled in debug mode.")
 # Compiler flags.
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -Wextra -Werror")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wno-unused-parameter -Wno-attributes") # TODO: remove
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -ggdb -fsanitize=${BUSTUB_SANITIZER} -fno-omit-frame-pointer -fno-optimize-sibling-calls")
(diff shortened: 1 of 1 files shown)


skyzh:
do you have any comparison on the debug information produced before and after this option is enabled? i.e., do you see more information in your debugger?

JinjieZhao:
Without -fstandalone-debug
p str -> error: summary string parsing error
image
With -fstandalone-debug
image

skyzh:
Hmm. This is weird as the debug info should be already embedded in the executable file? I assume the flag only helps generating some debug symbol files outside the executable.

JinjieZhao:
Yes, it is weird.

For gcc, this flag helps generating some debug symbol files outside the executable.
For clang, this flag adds more debug info in the executable. see https://clang.llvm.org/docs/UsersManual.html#controlling-debug-information

prashanthduvvada:
the '-fstandalone-debug' flag doesn't exist in g++. The g++ does all the prints that clang with '-fstandalone-debug' does by default. 

The fix isn't correct. It requires a check if it's g++ or clang.

CMakeLists.txt
```diff
@@ -116,7 +116,7 @@ message("${BUSTUB_SANITIZER} sanitizer will be enabled in debug mode.")
 # Compiler flags.
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -Wextra -Werror")
 set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wno-unused-parameter -Wno-attributes") # TODO: remove
-set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -ggdb -fsanitize=${BUSTUB_SANITIZER} -fno-omit-frame-pointer -fno-optimize-sibling-calls")
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -O0 -ggdb -fsanitize=${BUSTUB_SANITIZER} -fno-omit-frame-pointer -fno-optimize-sibling-calls -fstandalone-debug")
```
prashanthduvvada, line 119: Needs a check if it's clang compiler or not

Issue #625, opened 2023-10-15
bug: incorrect return type for ExtendibleHTableHeaderPage::GetDirectoryPageId


kenleung5e28:
The signature for `ExtendibleHTableHeaderPage::GetDirectoryPageId` is:
```cpp
auto GetDirectoryPageId(uint32_t directory_idx) const -> uint32_t;
```
Judging from the intent for this method, its return type should be `page_id_t` instead.

lidianzhong:
I think so, `page_id_t` type should be returned

PR #649, opened 2023-11-04
Update ExtendibleHTableHeaderPage::GetDirectoryPageId to page_id_t

jurplel:
Resolves #625

PR #683, opened 2024-01-12
change return type uint32_t -> page_id_t for func GetDirectoryPageId

dashjay:
An error type return by GetDirectoryPageId confused me for a long time, maybe there are also some who is not so familiar with CPP encountered this too. It return `INVALID_PAGE_ID(-1)` as uint32_t and get a super max one. Hope this can be fixed soon.

#625


dashjay:
@skyzh Hello MR. Chi PTLA~

Oh my good I finally found it.

skyzh:
dup w/ https://github.com/cmu-db/bustub/pull/649, ping current course staff for further actions, thanks :) @yliang412

PR #692, opened 2024-01-28
fix memory leak of vector type

skyzh:
backport from https://github.com/skyzh/bustub-vectordb/pull/1 @UnpureRationalist

---

Hi there,

Thanks for your excellent tutotial for vector database beginners. When I follow your tutorial and test my implementions using the commands [here](https://skyzh.github.io/write-you-a-vector-db/cpp-02-setup.html#testing), memory leak was reported by AddressSanitizer.

After debugging, I found the destructor of class `Value` was not implemented well for `VECTOR` type. So this PR fixed the bug by modifying destructor along with the copy constructor (The copy assignment operator do not need to change since it is pass by  value).

After fixing the bug, the test can passed correctly without memory leak.

### DIFF
--- a/src/type/value.cpp
+++ b/src/type/value.cpp
@@ -28,6 +28,7 @@ Value::Value(const Value &other) {
   value_ = other.value_;
   switch (type_id_) {
     case TypeId::VARCHAR:
+    case TypeId::VECTOR:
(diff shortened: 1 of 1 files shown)


yliang412 (approved):
LGTM!

PR #697, opened 2024-02-02
To match gradescope rules: Fix header file in buffer_pool_manager.h to pass gradescope p1&p2 test

shandianshan:
* Comment write_back_cache.h.
* Comment function call from write_back_cache.h.

Without changes below, gradescope will report error.
image


yliang412:
If you are using Fall 2023's Gradesocpe autograder, please see the [Fall 2023 Release](https://github.com/cmu-db/bustub/releases/tag/v20231227-2023fall). ** Please make sure your solution is private.**

PR #701, opened 2024-03-06
fix(p2): fix wrong definition of function

hnkjdaxzzq:
Modify the return value type of the `ExtendibleHTableHeaderPage::GetDirectoryPageId(uint32_t directory_idx) const -> uint32_t` function to `page_id_t`.

before:  Using this function get pageid requires type conversion.
After: The function return value conforms to the definition of the function name.

this function in `bustub/src/storage/page/extendible_htable_header_page.cpp`.


prashanthduvvada:
Fixed with #649

PR #712, opened 2024-04-12
Ignore format-truncation warning

radustoenescu:
With clang-17 compilation fails due to:

`src/type/timestamp_type.cpp:130:29: error: ‘%02d’ directive output may be truncated writing between 2 and 10 bytes into a region of size 5 [-Werror=format-truncation=]`

Until it gets a proper fix, ignore this warning.


prashanthduvvada:
Please check [CMakeLists.txt](https://github.com/cmu-db/bustub/blob/master/CMakeLists.txt#L42) for the same. It's suggested to use clang-14. I think it's due to the incompatibility with clang-17 that it may cause the issue.

Closing this as of now.

PR #717, opened 2024-07-16
fix third_party backward-cpp include dir path

luooofan:
## BUG description
The include dir path of the third-party library backward-cpp is set incorrectly. It should be `${PROJECT_SOURCE_DIR}/third_party/backward-cpp` rather than `${PROJECT_SOURCE_DIR}/third_party/backward-cpp/include`

if I want to use backward-cpp in the src directory, for example, include "backward.hpp" in the src/storage/page/extendible_htable_bucket_page.cpp, it will cause a compilation error: `fatal error: backward.hpp: No such file or directory`


prashanthduvvada (approved):
LGTM

PR #725, opened 2024-08-21
Bustub to BusTub - Naming consistency

prashanthduvvada:
Changed from Bustub - BusTub (naming consistency)

PR #726, opened 2024-08-21
CPP Standards

prashanthduvvada:
Changing from `std::string` to `std::filesystem::path` & `BustubInstance` to `BusTubInstance`


connortsui20 (approved):
LGTM, however we should probably rebase this so it's not dependent on #725. Or we can just merge #725 first.

PR #729, opened 2024-08-26
P0 Starter Code

prashanthduvvada:
Starter Code for P0.


cjtsui (approved):
It would be good if the doc comments had proper capitalization and punctuation, but other than that lgtm

connortsui20 (approved):
whoops wrong account

tre3p:
@prashanthduvvada maybe you can also add examples of cardinality calculations on the primer page? Cause you have examples of how data is laid out, but not how cardinality is calculated then based on this data. Thx in advance!

prashanthduvvada:
Okay ..that should fix most confusions. I was thinking about the same.

PR #730, opened 2024-08-27
Changes in Test case

prashanthduvvada:
Small change in Test cases

### DIFF
--- a/test/primer/hyperloglog_test.cpp
+++ b/test/primer/hyperloglog_test.cpp
@@ -1,4 +1,3 @@
-#include <iostream>
 #include <limits>
 #include <memory>
 #include <string>
@@ -12,7 +11,7 @@
 
 namespace bustub {
 
-TEST(HyperLogLogTestHyperLogLogTest, DISABLED_BasicTest1) {
+TEST(HyperLogLogTest, DISABLED_BasicTest1) {
   auto obj = HyperLogLog<std::string>(static_cast<uint64_t>(1));
 
   ASSERT_EQ(obj.GetCardinality(), 0);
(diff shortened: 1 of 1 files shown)


connortsui20 (approved):

test/primer/hyperloglog_test.cpp
```diff
@@ -22,7 +21,7 @@ TEST(HyperLogLogTestHyperLogLogTest, DISABLED_BasicTest1) {
   obj.ComputeCardinality();
   auto ans = obj.GetCardinality();
 
-  ASSERT_EQ(ans, 0);
+  ASSERT_EQ(ans, 2);
```
tre3p, line 24: Maybe I've implemented this incorrectly at the moment, but I keep getting 3 here for classic HLL(not Presto). Is this changes are final?

prashanthduvvada: Using std::floor?

tre3p: Yeah, for the first case, I'm retrieving 3.07984. When casted to size_t -> 3.
I can share my calculations if you need it

prashanthduvvada: Don't post anything here. Can lead to AIV. I think it may differ with the way you are calculating cardinality. 
Will post in website soon.
values should be final.

tre3p: Sure! Thanks, gonna wait

tre3p: @prashanthduvvada Hey! I figured out the problem, and I suggest you include that info on the course page. std::hash works differently on different CPU architectures (on ARM on x86 for sure), so that can cause issues for guys who going to try to implement primer on ARM MacBooks.

PR #732, opened 2024-08-30
p0-draft1

prashanthduvvada:
P0 - 1st Draft version.


src/include/primer/hyperloglog_presto.h
```diff
@@ -54,7 +58,20 @@ class HyperLogLogPresto {
...
+    /** @brief Convert into output stream */
+    std::ostringstream o_stream;
+
+    o_stream << val;  // output stream
+
+    /** @brief Output stream result. */
+    std::string res = o_stream.str();
+
+    return static_cast<hash_t>(std::stoul(res));
+  }
```
unw9527, line 74: I remember in the write-up, students are supposed to distinguish between `int` and `std::string` in the basic hyperloglog rather than the presto one. Did you change it?

src/include/primer/primer_hash.h
```diff
@@ -0,0 +1,45 @@
...
+class PrimerHashFunction {
+ private:
+  /** @brief hash value type */
+  using hash_t = uint64_t;
+
+  /** @brief seed constant*/
+  static const u_int32_t SEED = 0;
+
+ public:
+  auto inline GetHash(KeyType val) {
```
apavlo, line 21: Remove `inline`

src/include/primer/hyperloglog_presto.h
prashanthduvvada: Yes, changing the input type changes, where the string has to be converted to a c_str and then passed on to the function. Hence, removed that for now. Will update it in the writeup.

src/include/primer/primer_hash.h
```diff
@@ -0,0 +1,45 @@
...
+#include <memory>
+#include <sstream>
+#include <string>
+
+#include "murmur3/MurmurHash3.h"
+
+namespace bustub {
+
+template <typename KeyType>
+class PrimerHashFunction {
```
apavlo, line 12: Can we get rid of this and just use `HashUtil`?
https://github.com/cmu-db/bustub/blob/master/src/include/common/util/hash_util.h

apavlo (changes requested):
Do we really need `PrimerHashFunction`?

src/include/primer/primer_hash.h
prashanthduvvada: Done

test/primer/hyperloglog_test.cpp
```diff
@@ -133,15 +133,15 @@ TEST(HyperLogLogTest, DISABLED_BasicParallelTest) {
...
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Andy"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Connor"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("J-How"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Kunle"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Lan"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Prashanth"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Will"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Yash"); }));
-    threads2.emplace_back(std::thread([&]() { obj.AddElem("Yuanxin"); }));
+    threads2.emplace_back(std::thread([&]() { obj.AddElem("Andy Pavlo"); }));
```
apavlo, line 136: Remove last names.

test/primer/hyperloglog_test.cpp
```diff
@@ -154,10 +154,10 @@ TEST(HyperLogLogTest, DISABLED_BasicParallelTest) {
 }
 
 TEST(HyperLogLogTest, DISABLED_ParallelTest1) {
-  auto obj = HyperLogLog<std::string>(static_cast<uint64_t>(17));
+  auto obj = HyperLogLog<std::string>(static_cast<uint64_t>(14));
 
   std::vector<std::thread> threads1;
-  for (uint16_t i = 0; i < 10; i++) {
+  for (uint16_t i = 0; i < 1; i++) {
```
apavlo, line 160: Why only one thread?

apavlo (changes requested):
This is probably good enough for P0. Minor nits.

test/primer/hyperloglog_test.cpp
prashanthduvvada: Actually, I added it because I wanted the hash values to be distributed to different buckets. It was hitting the bucket with small strings.

test/primer/hyperloglog_test.cpp
prashanthduvvada: Done

test/primer/hyperloglog_test.cpp
prashanthduvvada: Done

src/include/primer/hyperloglog_presto.h
```diff
@@ -54,7 +54,17 @@ class HyperLogLogPresto {
...
    */
-  inline auto CalculateHash(T val) -> hash_t { return std::hash<T>{}(val); }
+  inline auto CalculateHash(T val) -> hash_t {
+    Value val_obj;
+    if constexpr (std::is_same<T, std::string>::value) {
+      val_obj = Value(VARCHAR, val);
+      return bustub::HashUtil::HashValue(&val_obj);
+    }
+    if constexpr (std::is_same<T, int64_t>::value) {
+      return static_cast<hash_t>(val);
```
vladtenlive, line 64: Why don't make it the same as`CalculateHash(T val) -> hash_t` in `hyperloglog.h`?
```

inline auto CalculateHash(T val) -> hash_t {
    Value val_obj;
    if constexpr (std::is_same<T, std::string>::value) {
      val_obj = Value(VARCHAR, val);
    } else {
      val_obj = Value(BIGINT, val);
    }
    return bustub::HashUtil::HashValue(&val_obj);
  }

```

and also there is an issue with MacOS tests in the current master branch

prashanthduvvada: For testing special scenarios! Hence, doing the same.

test/primer/hyperloglog_test.cpp
```diff
@@ -84,39 +83,40 @@ TEST(HyperLogLogTest, DISABLED_BasicTest2) {
...
 }
 
 TEST(HyperLogLogTest, DISABLED_EdgeTest1) {
-  auto obj1 = HyperLogLog<int64_t>(-2);
+  auto obj1 = HyperLogLog<int64_t>(static_cast<int16_t>(-2));
   obj1.ComputeCardinality();
   ASSERT_EQ(obj1.GetCardinality(), 0);
 }
 
 TEST(HyperLogLogTest, DISABLED_Edgetest2) {
```
connortsui20, line 104: This needs to change to `EdgeTest2` with a capital, I changed it in the private version

connortsui20 (changes requested):

connortsui20 (approved):

src/include/primer/hyperloglog_presto.h
prashanthduvvada: The error is fixed now...can clone the latest changes

PR #731, opened 2024-08-28
change LRU-K `Evict()` signature

connortsui20:
This PR changes the signature of the `Evict` method on the LRU-K replacer to use a `std::optional` return type containing the frame instead of writing it into a pointer.

PR #734, opened 2024-09-01
Buffer Pool Manager (Project 1) Refactor

connortsui20:
This PR will be the main PR for refactoring the buffer pool manager, which is the entirety of Project 1.

The overarching goal of this refactoring is to make the buffer pool manager API both thread-safe and type-safe. This means that use of raw pointers both internally and externally should be removed, and that thread-safe page guards should be the _only_ way to interact with page data.

By doing this, we heavily limit the opportunities for misuse of the buffer pool manager API, which will in turn make the implementation and debugging of future projects (which will be much harder) easier.

Additionally, there will be implementation changes for the reference solution that will prefer indexing into arrays over raw pointers, limiting the opportunities for segfaults and memory corruption.

Here are the steps required for the refactor:
- [X] Make any modifications to the LRU-K replacer and disk scheduler (#731)
- [X] Change the BPM API in the public repo (d1fc14f02d1e4f70d3679d86f0ccc20b32b80feb)
    - [x] Propagate the BPM API changes in the public repo (change occurrences of old methods to new ones)
    - [x] Update tests that require OOM error / require old raw pointer API
- [x] Update the writeup [in progress]
- [x] Make all private repo changes (see private PR for more information [in progress])
- [x] Update all test cases
- [x] Update the starter code documentation
- [x] [Potentially more steps necessary + review]


src/include/common/config.h
```diff
@@ -33,12 +33,14 @@ extern std::chrono::duration<int64_t> log_timeout;
...
-static constexpr int HEADER_PAGE_ID = 0;                                             // the header page id
 static constexpr int BUSTUB_PAGE_SIZE = 4096;                                        // size of a data page in byte
 static constexpr int BUFFER_POOL_SIZE = 10;                                          // size of buffer pool
 static constexpr int LOG_BUFFER_SIZE = ((BUFFER_POOL_SIZE + 1) * BUSTUB_PAGE_SIZE);  // size of a log buffer in byte
 static constexpr int BUCKET_SIZE = 50;                                               // size of extendible hash bucket
-static constexpr int LRUK_REPLACER_K = 10;  // lookback window for lru-k replacer
+static constexpr int LRUK_REPLACER_K = 10;  // default lookback window for lru-k replacer
+
+// TODO(cjtsui) what is this HEADER_PAGE_ID???
+static constexpr int HEADER_PAGE_ID = 0;  // the header page id
```
xx01cyx, line 43: Seems that this is only used in tests... Should consider refactoring

src/buffer/buffer_pool_manager.cpp
```diff
@@ -36,32 +48,171 @@ BufferPoolManager::BufferPoolManager(size_t pool_size, DiskManager *disk_manager
...
+ * @brief Fetch the requested page from the buffer pool. Return nullptr if page_id needs to be fetched from the disk
+ * but all frames are currently in use and not evictable (in another word, pinned).
+ *
+ * First search for page_id in the buffer pool. If not found, pick a replacement frame from either the free list or
+ * the replacer (always find from the free list first), read the page from disk by scheduling a read DiskRequest with
+ * disk_scheduler_->Schedule(), and replace the old page in the frame. Similar to NewPage(), if the old page is dirty,
+ * you need to write it back to disk and update the metadata of the new page
+ *
+ * In addition, remember to disable eviction and record the access history of the frame like you did for NewPage().
+ *
```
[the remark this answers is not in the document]
connortsui20: Yes, I just didn't modify the old documentation because I want to remove this function entirely

src/include/common/config.h
connortsui20: #736

src/include/storage/disk/disk_manager_memory.h
```diff
@@ -67,6 +74,11 @@ class DiskManagerUnlimitedMemory : public DiskManager {
  public:
   DiskManagerUnlimitedMemory() { std::fill(recent_access_.begin(), recent_access_.end(), -1); }
 
+  /**
+   * This function should increase the disk space, but since this is memory we just resize the vector.
+   */
+  void IncreaseDiskSpace(size_t pages) override { data_.resize(pages + 1, std::make_shared<ProtectedPage>()); }
```
connortsui20, line 80: I think this is wrong. It has not effect on P1 but it still needs to be fixed.

src/include/storage/page/page.h
```diff
@@ -14,54 +14,62 @@
...
 #include "common/config.h"
 #include "common/rwlatch.h"
 
 namespace bustub {
 
 /**
+ * @brief
+ *
+ * TODO(2024 tas) This entire class needs to be updated.
+ *
```
connortsui20, line 28: We should rip out this entire file.

connortsui20:
We need to also decide if we want to put the detailed implementation hints in the writeup (website) or keep them in the doc comments. Personally, I feel that documentation should always go right next to where the function implementation is, but I also understand that I made the implementation hints really long and detailed.

test/storage/page_guard_test.cpp
```diff
@@ -22,32 +22,175 @@
 
 namespace bustub {
 
-// NOLINTNEXTLINE
-TEST(PageGuardTest, DISABLED_SampleTest) {
-  const size_t buffer_pool_size = 5;
-  const size_t k = 2;
+const size_t FRAMES = 10;
```
unw9527, line 25: Consider renaming this? This is actually still size, isn't it?

src/storage/disk/disk_manager.cpp
```diff
@@ -26,6 +26,9 @@ namespace bustub {
 
 static char *buffer_used;
 
+/** @brief The default size of the database file. */
+static const size_t DEFAULT_DB_IO_SIZE = 16;
+
```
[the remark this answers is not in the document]
connortsui20: No reason. It doesn't really matter as long as its small

test/storage/page_guard_test.cpp
connortsui20: # of buffer frames is more accurate.

src/include/storage/page/page.h
connortsui20: Other parts of the system still rely on this file and we haven't ripped them out yet. We'll update this once we start making changes for P2

unw9527 (approved):

PR #739, opened 2024-09-07
extend support for Ubuntu `24.04`

rolandvarga:
Tiny change to support current Ubuntu LTS

### DIFF
--- a/build_support/packages.sh
+++ b/build_support/packages.sh
@@ -45,6 +45,7 @@ install() {
         18.04) install_linux ;;
         20.04) install_linux ;;
         22.04) install_linux ;;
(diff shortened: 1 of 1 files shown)


connortsui20:
This seems fine to me, though I'm unable to test this myself.

skyzh:
bumping distro version requires students to upgrade their LLVM version; this is better done at the beginning of the semester :(

rolandvarga:
> bumping distro version requires students to upgrade their LLVM version; this is better done at the beginning of the semester :(

This change doesn't force anyone to have to update their dev environment. It's simply a QOL change for those who run the latest version of Ubuntu.

PR #741, opened 2024-09-15
Add a Project #1 Deadlock Test

connortsui20:
This test adds a deadlock test into the public project 1 tests.

We need to add this otherwise students may run into problems in future projects.


apavlo (approved):
LGTM

PR #745, opened 2024-09-18
Add `EvictableTest`

connortsui20:
This test checks if an implementation has a pin count / evictability mismatch bug.


J-HowHuang (approved):
lgtm

PR #753, opened 2024-09-28
Add new frame latching test

connortsui20:
This test ensure that people actually use frame latches, because it is possible to pass a lot of the tests if you completely disregard locking.


unw9527 (approved):
LGTM. Have you tested if this new test can fail those who do not use latches?

cjtsui:
> LGTM. Have you tested if this new test can fail those who do not use latches?

Yep. If they don't use latches at all this immediately fails.

connortsui20:
> LGTM. Have you tested if this new test can fail those who do not use latches?

Yep, if they do not use latches at all this immediately fails.

PR #761, opened 2024-10-18
feat(f24-p3): add new mock tables

xx01cyx:
Add new mock tables for fall 2024. Introduce mock tables for testing external merge sort.

### DIFF
--- a/src/execution/mock_scan_executor.cpp
+++ b/src/execution/mock_scan_executor.cpp
@@ -407,7 +407,7 @@ auto GetFunctionOf(const MockScanPlanNode *plan) -> std::function<Tuple(size_t)>
       return Tuple{values, &plan->OutputSchema()};
     };
   }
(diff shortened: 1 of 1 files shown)


src/execution/mock_scan_executor.cpp
```diff
@@ -374,6 +407,16 @@ auto GetFunctionOf(const MockScanPlanNode *plan) -> std::function<Tuple(size_t)>
       return Tuple{values, &plan->OutputSchema()};
     };
   }
+  // select * from __mock_external_merge_sort_input order by v2 desc, v3 asc limit 100;
```
prashanthduvvada, line 410: Is this comment required?

xx01cyx: good catch. I'll remove it.

prashanthduvvada (approved):
I've added one comment, but it's not a blocker so I'm approving it. LG

PR #770, opened 2024-11-08
f24-p3: update test for index scan

xx01cyx:


test/sql/p3.05-index-scan-btree.slt
```diff
@@ -0,0 +1,466 @@
+// :bustub-keep-private:
```
connortsui20, line 1: I think this shouldn't be here?

test/sql/p3.05-index-scan-hash.slt
J-HowHuang, line 1: Remove the p3.05 prefix in the file name of this test case? Since it's not relevant and should not be included in p3 for now

J-HowHuang (approved):
lgtm

PR #773, opened 2024-11-17
fix typo

Wangch29:
Fix typo, it should be topn_per_group_executor.cpp instead of topn_executor.cpp.


connortsui20 (approved):
thanks for the catch!

PR #786, opened 2024-12-10
Add stream-style logging macros

dentiny:
Problem statement:
- Existing logging and assertion is kind of awkward, for example, it doesn't accept stream-style operation, meanwhile it uses string as the macro parameter, which means we have to concatenate the message beforehand.
  + (minor) It might hurts performance, I see a lot of the message construction is made by `operator+`, which could create temporary strings
  + (major) it reduces logging util usability

In this PR, I propose to add a new macro, which combines both assertion and logging with stream style operation.

### DIFF
--- a/src/include/common/macros.h
+++ b/src/include/common/macros.h
@@ -14,11 +14,45 @@
 
 #include <cassert>
 #include <exception>
+#include <sstream>
 #include <stdexcept>
 
 namespace bustub {
 
 #define BUSTUB_ASSERT(expr, message) assert((expr) && (message))
+
+namespace internal {
+
+// An internal stream, used to take C++ stream-style parameters for display, and exit the program with `std::abort`.
+class LogFatalStream {
+ public:
+  LogFatalStream(const char *file, int line) : file_(file), line_(line) {}
+
+  ~LogFatalStream() {
+    std::cerr << file_ << ":" << line_ << ": " << log_stream_.str() << std::endl;
+    std::abort();
+  }
+
+  template <typename T>
+  LogFatalStream &operator<<(const T &val) {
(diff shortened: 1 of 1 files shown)


src/include/common/macros.h
```diff
@@ -14,12 +14,46 @@
...
+  int line_;
+  std::ostringstream log_stream_;
+};
+
+}  // namespace internal
+
+// A macro which checks `expr` value and performs assert.
+// Different from `BUSTUB_ASSERT`, it takes stream-style parameters.
+#define BUSTUB_ASSERT_AND_LOG(expr)             \
+  if (bool val = (expr); !val) LogFatalStream { \
```
dentiny, line 53: If you think necessary, I could add a `Unlikely` function, which improves branch prediction, since I suspect all assertions sits on critical path.

apavlo:
@dentiny Thanks for sending this. We will take a look at it. It might make more sense to use a third-party header-only logger.

dentiny:
Thanks @apavlo for taking a look! It would be nice to support `CHECK_EQ`-like syntax, which
- Allows comparison logic
- Prints out value for lhs and rhs if comparison logic fails
- Code location (i.e. filename and line number) is included in the error message

For this PR, I mostly target at code simplicity w/o extra deps :)

dentiny:
Curious if there're any updates on this PR? :eyes:

apavlo:
@dentiny End of semester crunch. We will look at this soon.

connortsui20 (approved):
This looks fine to me. Sorry for not reviewing this last semester! I guess this will be up to the future TAs to actually merge but LGTM

PR #793, opened 2025-01-06
Move all function comments to source files

xueqili02:
move all function comments from headers to source files in doxygen style.


connortsui20:
Similar to private repo, if CI passes we should just merge this and we can slowly clean up things over time to keep this PR simple.

PR #799, opened 2025-01-20
fix(f24 p1): update Evict function doc for `std::optional` return type

takaebato:
Fixes the Evict function documentation for the `std::optional` return type in Project 1 of Fall 2024.


connortsui20 (approved):
LGTM

PR #800, opened 2025-01-20
p1 (bpm) flush api changes

rmboyce:
Change buffer pool manager flush API to be more understandable:

1. Added flush methods on read and write page guards that will always flush a consistent state to disk (due to the guard locking the page).
2. Added safe and unsafe flush methods to the buffer pool manager that lock the pages and don't respectively. The unsafe ones will be faster due to not having to lock pages, but they may flush an inconsistent state to disk if the page is modified while it is flushed. Note that previously, the buffer pool manager only had unsafe flush methods.


connortsui20 (approved):
Nice, this was sorely needed last semester and we didn't have time to get to it!

MrWhitezz (commented):
BPM flush api and Page Guard api updated correctly.

PR #801, opened 2025-01-24
Support clang 18

rmboyce:
Allow BusTub to build on Clang 18 (I know Clang 14 is officially recommended, but I thought it would still be nice if it built on Clang 18).

Changes:
- Update version of third-party library fmt, as it was causing compiler warnings (that get converted to errors by the compile flags)
- Include fmt/ranges.h in files that needed fmt::join, as this was split off in the fmt library
- Miscellaneous small changes to files that gave compile errors on Clang 18


yliang412 (approved):
LGTM. One other thing we could do is to include clang-18 in the CI. But we can do it maybe in another PR.

PR #802, opened 2025-01-27
Fix p1 (bpm) new page documentation

rmboyce:
Remove note saying to use `DiskScheduler::IncreaseDiskSpace`, as disk space is maintained internally by the disk manager now.


MrWhitezz (approved):
Reasonable comment update

Issue #821, opened 2025-06-11
build fail on Ubuntu 24.04 LTS


yoyedmundyoy:
Is this not supported yet?

apavlo:
We will add this for Fall 2025.

PR #827, opened 2025-08-24
Added support for Ubuntu 24.04. Dropped Ubuntu 18.04

apavlo:
This bumps up support to Ubuntu 24.04 and MacOS 14 + 15.
We are switching to clang-15 (since this is the latest version still supported on Ubuntu 22.04).

#821

### DIFF
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -37,14 +37,14 @@ jobs:
           - {
               name: "macOS 14 (Sonoma) AppleClang",
               os: macos-14,
-              format: "/usr/local/opt/llvm@15/bin/clang-format",
-              tidy: "/usr/local/opt/llvm@15/bin/clang-tidy",
+              format: "/opt/homebrew/opt/llvm@15/bin/clang-format",
+              tidy: "/opt/homebrew/opt/llvm@15/bin/clang-tidy",
             }
           - {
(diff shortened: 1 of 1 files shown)


src/storage/page/table_page.cpp
```diff
@@ -55,7 +55,7 @@ auto TablePage::GetNextTupleOffset(const TupleMeta &meta, const Tuple &tuple) co
  */
 auto TablePage::InsertTuple(const TupleMeta &meta, const Tuple &tuple) -> std::optional<uint16_t> {
   auto tuple_offset = GetNextTupleOffset(meta, tuple);
-  if (tuple_offset == std::nullopt) {
+  if (tuple_offset == std::nullopt || !tuple_offset.has_value()) {
```
songwdfu, line 58: [minor] Is this necessary?

songwdfu:
```suggestion
  if (!tuple_offset.has_value()) {
```

songwdfu (commented):
one minor, otherwise LGTM

src/storage/page/table_page.cpp
apavlo: I think this fix came from `clang-tidy-15 --fix --fix-errors ...`

songwdfu (approved):

PR #830, opened 2025-08-26
f25-p0 fix: add back seeded HashFunction util, mod contention ratio test

songwdfu:
Added back missing seeded HashFunction utils

Modified contention ratio test to also check for correctness


rayhhome (approved):
Lgtm!

PR #835, opened 2025-08-30
Chore: Clean up some headers

connortsui20:
Ideally I would clean up more than this, but I unfortunately do not have time


connortsui20:
hmm evidently they do matter for MacOS 15 (which is weird since I'm developing off of MacOS 15 and I don't have this problem)

PR #844, opened 2025-09-25
P2 update F2025

quantumish:
Added optimistic latching tests + leaf page tombstoning.


src/include/buffer/traced_buffer_pool_manager.h
```diff
@@ -0,0 +1,33 @@
...
+ private:
+  size_t reads_ = 0;
+  size_t writes_ = 0;
+  BufferPoolManager *bpm_;
+
+ public:
+  explicit TracedBufferPoolManager(BufferPoolManager *bpm) : bpm_{bpm} {}
+
+  auto WritePage(page_id_t page_id, AccessType access_type = AccessType::Unknown) -> WritePageGuard {
+    writes_++;
```
skyzh, line 17: sounds like we need to make it atomic?

PR #848, opened 2025-10-04
Remove ContentionTest from test folder

quantumish:
Removes the contention test from the testing folder to prevent students from thinking it's part of this semester's assignment.


rayhhome (approved):
looks good

Issue #834, opened 2025-08-29
Build failure on MacOs


varunu28:
I am facing the following build failure after running `make`.

```
In file included from /Users/varunu28/projects/bustub/third_party/googletest/googlemock/src/gmock-all.cc:43:
/Users/varunu28/projects/bustub/third_party/googletest/googlemock/src/gmock-internal-utils.cc:185:36: error: too few arguments to function call, expected 2, have 1
  184 |               << ::testing::internal::GetCurrentOsStackTraceExceptTop(
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  185 |                      actual_to_skip);
      |                                    ^
/usr/local/include/gtest/internal/gtest-internal.h:839:24: note: 'GetCurrentOsStackTraceExceptTop' declared here
  839 | GTEST_API_ std::string GetCurrentOsStackTraceExceptTop(UnitTest* unit_test,
      |                        ^                               ~~~~~~~~~~~~~~~~~~~~
  840 |                                                        int skip_count);
      |                                                        ~~~~~~~~~~~~~~
1 error generated.
```

The above error happens after I successfully run `cmake ..`. 

Output for `g++ --version`
```
g++ --version
Apple clang version 17.0.0 (clang-1700.0.13.5)
Target: arm64-apple-darwin24.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

apavlo:
@varunu28 Can you send a PR with the fix?

PR #855, opened 2025-10-30
(fix) Fixing build issue for googletest on Mac

varunu28:
### Heads Up
Please note that this PR was created by someone who has almost no knowledge of C++ build system. So if you think something doesn't makes sense then there is a pretty good chance that you are right. I have almost no clue of the CMake build system & I am happy to learn the ropes here. Also for full transparency, I have leveraged LLM tools to understand the underlying root cause & triage this issue. 

#### Why this change?
This change tries to address #834 . 

#### Actual code change
The issue was happening as system headers were given a priority over the bundled third party dependencies due to usage of `SYSTEM` keyword. The fix removes the `SYSTEM` keyword from `CMakeLists.txt` from both `googletest` dependencies & also adds bundled `googletest` include directories with `BEFORE` keyword in order for CMake to include these paths in the beginning of dependency scan.

#### Testing
`make` finishes successfully
image

Also invoked test & it failed(as expected) which verifies that the test library is working as intended
Screenshot 2025-10-30 at 9 55 07 AM

Screenshot 2025-10-31 at 5 37 53 AM

### DIFF
--- a/src/concurrency/transaction_manager.cpp
+++ b/src/concurrency/transaction_manager.cpp
@@ -47,7 +47,7 @@ auto TransactionManager::Begin(IsolationLevel isolation_level) -> Transaction * 
   auto *txn_ref = txn.get();
   txn_map_.insert(std::make_pair(txn_id, std::move(txn)));
 
-  // TODO(fall2023): set the timestamps here. Watermark updated below.
+  // TODO(P4): set the timestamps here. Watermark updated below.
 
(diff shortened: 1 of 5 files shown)


varunu28:
@apavlo I have updated the check to be applied only when googletest dependency is present in the system.

varunu28:
@apavlo Can you please re-trigger the build workflow to verify the latest commit? Thanks :)

PR #893, opened 2026-08-31
test: cover serializable validation after transaction GC

wodex1nhaoIeng:
## Summary

This PR adds a disabled Project 4 regression test covering the interaction between serializable validation and transaction garbage collection.

## Motivation

The existing tests cover serializable validation and garbage collection separately, but do not test what happens when garbage collection runs before an older serializable transaction commits.

This matters because a newly inserted tuple has no previous version and therefore may not create an `UndoLog`. After the inserting transaction commits and garbage collection runs, the system must still preserve enough information for an older serializable transaction to detect that the insertion changed the result of its earlier predicate scan.

Without this coverage, an implementation may pass both the existing serializable and garbage-collection tests while incorrectly accepting a non-serializable execution.

## Test Scenario

1. A serializable transaction scans an empty predicate (`a > 20`).
2. Another transaction inserts a tuple matching that predicate and commits.
3. The first transaction performs a non-conflicting write.
4. Transaction garbage collection runs.
5. The first transaction attempts to commit.

The final commit must fail because the concurrent insertion introduces a phantom that changes the result of the earlier predicate scan.

The test is disabled by default, consistent with the existing Project 4 transaction tests.

### DIFF
--- a/test/txn/txn_abort_serializable_test.cpp
+++ b/test/txn/txn_abort_serializable_test.cpp
@@ -99,6 +99,26 @@ TEST(TxnSerializableTest, DISABLED_ConcurrentSerializableTest) {  // NOLINT
   }
 }
 
+TEST(TxnSerializableTest, DISABLED_GcMustNotEraseCommittedWriteSetNeededForValidation) {  // NOLINT
+  auto bustub = std::make_unique<BusTubInstance>();
+  Execute(*bustub, "CREATE TABLE maintable(a int, b int)");
+
+  auto txn1 = BeginTxnSerializable(*bustub, "txn1");
+  WithTxn(txn1,
+          QueryShowResult(*bustub, _var, _txn, "SELECT * FROM maintable WHERE a > 20", IntResult{}));
+
+  auto txn2 = BeginTxn(*bustub, "txn2");
+  WithTxn(txn2, ExecuteTxn(*bustub, _var, _txn, "INSERT INTO maintable VALUES (30, 1)"));
+  const auto txn2_id = txn2->GetTransactionId();
+  WithTxn(txn2, CommitTxn(*bustub, _var, _txn));
+
+  WithTxn(txn1, ExecuteTxn(*bustub, _var, _txn, "INSERT INTO maintable VALUES (10, 1)"));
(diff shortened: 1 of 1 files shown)


wodex1nhaoIeng:
BTW you guys may also want to update it to gradescope cuz I passed the gradescope last semester but I just found this bug when I reviewed my code last night and apparently it shouldn't get passed

jmueducn:
sounds reasonable

0legmak:
I don't think "Task 3.5 Stop-the-world Garbage Collection" is supposed to support transactions running at the serializable isolation level. Here is a project write-up quote:

> You will first need to implement the SNAPSHOT ISOLATION isolation level for all of your transactions. You will then expand your concurrency control protocol to support the SERIALIZABLE isolation level in [Task #4.4](https://15445.courses.cs.cmu.edu/fall2025/project4/#task4.4).

You may want to implement garbage collection for serializable transactions in the leaderboard benchmark, but it should be a separate implementation, as stop-the-world garbage collection is not used there.

Also, I think you are not supposed to mix different isolation levels in one test. Quote:

> In every test case, all transactions in the test case run at the same isolation level.

wodex1nhaoIeng:
Thank you for the clarification. I may be misunderstanding your point: do you mean that garbage collection should never be invoked while a Serializable transaction is active?

I did not add any Serializable tests to Task 3.5. I added the GC-related case to the Task 4.4 Serializable test file because it exposes implementations that incorrectly remove committed transaction metadata needed for serializability validation. In some implementation, GC removes that metadata, causing an invalid implementation to incorrectly pass,.

You are correct about not mixing isolation levels within one test case. I have modified the test so that all transactions in it now use the same isolation level.

0legmak:
Stop-the-world garbage collection from task 3.5 should never be invoked with serializable transactions because students aren't required to implement support for serializable transactions in this task (only task 4.4 and the leaderboard require support for serializable transactions). Students may optionally decide to implement fine-grained garbage collection for the leaderboard benchmark, which must support serializable transactions, but this has to be a different implementation of garbage collection; in other words, they are not required to add support for serializable transactions to stop-the-world GC, which is not used by the leaderboard test.

